PR #58, opened 2015-01-28
fix test path

frankleng:
@koistya Jest could not find the sample test case.
per https://facebook.github.io/jest/
the folder should be named '__tests__' instead


koistya:
@frankleng good catch!

PR #60, opened 2015-02-08
Use 6to5 rather than jstransform

### LINKED ISSUE #53: Use 6to5 transform instead of ReactTools transform
6to5 handles JSX just fine, but it also gives a lot more ES6 features.

goatslacker:
This closes #53

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -53,19 +53,8 @@ var pkgs = (function() {
   return pkgs;
 }());
 
-// Configure JSX Harmony transform in order to be able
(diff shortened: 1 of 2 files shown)


koistya:
Thanks for the PR! I have replaced `jsx-loader` with `6to5-loader` in [webpack.config.js](https://github.com/kriasoft/react-starter-kit/blob/9f490661270886c025f18ff1cd2fb6c1941b0059/config/webpack.js#L88-L90) and also it should work for both client-side and server-side bundles (used for server-side rendering).

PR #62, opened 2015-02-09
Stops browsersync from opening

goatslacker:
browsersync insists on opening localhost:3000 which is not where the app is at.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -142,7 +142,9 @@ gulp.task('serve', function(cb) {
         https: false,
         // Informs browser-sync to proxy our Express app which would run
         // at the following location
-        proxy: 'http://localhost:5000'
+        proxy: 'http://localhost:5000',
+
+        open: false
       });
 
       process.on('exit', function () {


goatslacker:
Build failure is unrelated to commit

koistya:
@goatslacker the `open: false` flag tells `BrwoserSync` not to open an instance of a browser after its lunch. Maybe we should keep it by default (which is `open: true`)? See: http://www.browsersync.io/docs/options/#option-open

koistya:
@goatslacker also note that [BrowserSync](http://www.browsersync.io/) is currently configured to be used as a proxy a proxy, it creates `localhost:3000` server over our `express.js`-based `localhost:5000` server.

goatslacker:
Hmm, that's not working for me. When I run `gulp` off a fresh clone and npm install it opens up `localhost:3000` and times out. If I visit `localhost:5000` the site works as expected.

koistya:
@goatslacker when you do `npm install`, do you see any error messages? Maybe browser-sync wasn't properly installed? It depends on this thing https://github.com/TooTallNate/node-gyp/

goatslacker:
I'm normally on iojs v1.1.0 (which it errors) but I also tried node v0.12.0 (also errors) and node v0.10.36 (doesn't error)

koistya:
Hm... does it print any error messages or warnings to the console when you run `gulp`?

goatslacker:
No it runs fine, which is weird.

screen shot 2015-02-08 at 11 36 30 pm

But then localhost:3000 will hang forever

koistya:
It seems like `Express.js` server starts before BrowserSync's `localhost:3000`, maybe that's the problem... I will check `gulpfile.js/serve` for a possible cause of it. On my machine `localhost:5000` is launched first, then goes `localhost:3000`..

goatslacker:
Adding a timeout works, but that's a shitty solution.

koistya:
Yeah, perhaps there is a tiny bug in the `nodemon` startup script:

https://github.com/kriasoft/react-starter-kit/blob/6b4520ac1cb8d40df3aad3989f38f18730da1f3d/gulpfile.js#L124-L161

goatslacker:
Here's a radical idea:

Create a `gulp serve` script which just starts the server and doesn't use browser-sync.
Create a `gulp hot` script which starts react-hot-loader.

And then make browser-sync optional rather than the default.

??

koistya:
Yep, something like that.. for example we can move `browser-sync` from the `gulp serve` into a newly created `gulp watch` and maybe plug `react-hot-realod` into this same `gulp watch` task alongside with `browser-sync`? Or, yeah, create two additional tasks `gulp watch` for `react-hot-reload` and `gulp sync` for `browser-sync` which both would depend on `gulp serve` (the one starting `express.js` server).

goatslacker:
`gulp watch`, `gulp sync`, and `gulp serve` sound fine.

koistya:
..or, maybe have `react-hot-reload` inside the `gulp serve` task and just move `browser-sync` into its own `gulp sync` task.

goatslacker:
New PR coming up.

PR #64, opened 2015-02-09
Fixes jest dir

goatslacker:
This PR fixes the build.


package.json
```diff
@@ -64,7 +64,7 @@
     "webpack-dev-server": "^1.7.0"
   },
   "jest": {
-    "rootDir": "src",
+    "rootDir": "./",
```
koistya: If you set [jest](http://facebook.github.io/jest/)'s `rootDir` to `./` it will also search for tests inside the `./node_modules` folder.

goatslacker: This is related to the build failures, it's looking for jest inside `src/node_modules` where it doesn't exist.

goatslacker:
Rebased.

koistya:
While `var logo = require('./logo-small.png');` (line 13) can be removed, I'm trying to figure if if there is a way to make [jest](http://facebook.github.io/jest/) work with stuff like ``...

goatslacker:
Maybe I'm not webpack convinced yet so I'd argue against requiring images directly for the following reasons:

* It adds complexity
* It increases the bundle size by a lot (especially if you have larger images)
* It increases the markup size, a web request for the image is preferable

The way to fix jest would be to pre-process these images to return something blank. But again, back to my first point, complexity.

goatslacker:
Awesome that the build is fixed! I'd still like to propose this change based on the points I noted above.

koistya:
@goatslacker how do you propose to change it? How do you keep the image in the component's folder and yet reference it from HTML, so that URL continues to work after compilation.

goatslacker:
We have a static images folder and we use the uri to reference it, http will take care of the rest.

PR #65, opened 2015-02-09
Splits browser sync and serve

goatslacker:
Ref: #62

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -115,14 +115,11 @@ gulp.task('build', ['clean'], function(cb) {
 
 // Launch a lightweight HTTP Server
 gulp.task('serve', function(cb) {
-
   var nodemon = require('nodemon');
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -132,34 +129,43 @@ gulp.task('serve', function(cb) {
...
       gulp.watch(src.styles, ['styles']);
-      gulp.watch(DEST + '/**/*.*', function (file) {
-        browserSync.reload(path.relative(__dirname, file.path));
-      });
       cb();
     });
   });
 });
 
+gulp.task('sync', function() {
```
koistya: Perhaps it should depend on `serve`? `gulp.task('sync', ['serve'], function() { ... });`

goatslacker: Good call. I'll remove the runSequence call.

PR #66, opened 2015-02-09
Updated Readme.md to reflect new dir structure

sdiaz:


koistya:
The `styles` folder is going to be removed in favor of CSS per component.. How about this?

```
├── /src/                       # The source code of the application
│   ├── /actions/               # Action creators that allow to trigger a dispatch to stores
│   ├── /assets/                # Static files which are copied to ./build on compile
│   ├── /components/            # React components
│   ├── /constants/             # Enumerations used in action creators and stores
│   ├── /content/               # Web page content (HTML, Markdown, Jade, images etc.)
│   ├── /core/                  # Core components (Flux dispatcher, base classes, utilities)
│   ├── /stores/                # Stores contain the application state and logic
│   ├── /templates/             # HTML templates (for server-side rendering, emails etc.)
│   ├── /app.js                 # Client-side startup script
│   └── /server.js              # Server-side startup script
```

Issue #68, opened 2015-02-09
Old content is served


zerkms:
When I change any `content/*.jade` file the old content is served until `gulp` server restart.

Steps to reproduce:

1. Run `gulp`
2. Change `src/content/index.jade` somehow (Eg: add some characters to any header)
3. Notice there is

        [22:23:00] Starting 'assets'...
        [22:23:00] 'assets' all files 1.03 kB
        [22:23:00] Finished 'assets' after 31 ms
        [QC] File changed: build/content/index.jade

4. Check that `build/content/index.jade` actually changed

Expected result: the rendered view in the browser should change

Actual result: it does not. Even refreshing the page does not help

PS: Both `http://localhost:5000/` and `http://localhost:3000/` show the old content

The same is reproducible with `gulp serve`. And running an `rs` command (or a complete `nodemon` server restart) followed by a page refresh actually leads to the updated contents.

Meanwhile changing styles works as expected (in both cases: with `serve` and live reload)

Summary: everything from `styles` is updated fine, `content` and `templates` are not updated though.

Reproducible in both `v0.3.0` and current master `b9b42223d568`

Any ideas on why it could be like that?

frankleng:
@koistya  it's rendered and cached in memory. was this intended?

zerkms:
@frankleng I would be surprised if it was: you need to restart a server manually with `rs` every time you changed something.

frankleng:
@zerkms  true for dev. but it makes sense on production to have content live in RAM.

koistya:
> is building pages in JSX not best practice?

You want to use React/JSX mostly for UI elements, but for the actual website content the more traditional format is more appropriate (plain text or HTML, Markdown, Jade etc.). Consider for example a huge privacy policy page, or an article, blog post, that kind of content usually comes from a database, but for simplicity it can be stored in files (.html, .md, .json etc.)

koistya:
@zerkms, @frankleng check it out, the browser now refreshes when you change content pages (`.jade` files).

PR #69, opened 2015-02-10
fix .jade templates not reloading upon change

frankleng:
my hack for solving #68 
@koistya  please advise

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -15,7 +15,7 @@ import express from 'express';
 import React from 'react';
 
 // Set global variables
-global.__DEV__ = process.env.NODE_ENV == 'development';
+global.__DEV__ = (process.env.NODE_ENV == 'development');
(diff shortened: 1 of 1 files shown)


src/server.js
```diff
@@ -15,7 +15,7 @@ import express from 'express';
 import React from 'react';
 
 // Set global variables
-global.__DEV__ = process.env.NODE_ENV == 'development';
+global.__DEV__ = (process.env.NODE_ENV == 'development');
```
frankleng: making a bit clearer that you are assigning a boolean

zerkms:
Does it also address changes in react components? Or that should be done using `react-hot-loader` instead?

frankleng:
@zerkms @koistya react-hot-loader would be ideal. right now the server just restarts when a component is changed. this PR does not address that

Issue #71, opened 2015-02-22
Use eslint instead of jshint


dmnd:
Eslint supports JSX and some ES6 features. That will allow you to get rid of all those `/* jshint ignore:start */`. 

Note that right now ES6 imports are not yet implemented by eslint, but they should be coming soon.

koistya:
:+1:

sdiaz:
:+1:

ivanoats:
Until ESLint supports ES6 modules, you can also use JSCS and Esprima-fb to lint JSX / ES6 code. See config in https://youtrack.jetbrains.com/issue/WEB-13511

PR #74, opened 2015-02-28
Use ESLint Loader

nason:
Hey thanks for this project!

#71 switched JSHint out for ESLint, but did not update the webpack loader. The `gulp bundle` task was failing silently. This change restores bundling by using `eslint-loader`.


koistya:
@nason cool, thanks!

PR #80, opened 2015-03-18
Fix output.publicPath in wepback.config.js

abramz:
Changed output.publicPath from './build/' to './'

Images referenced in CSS by the 'url' statement were improperly getting referenced with '/build/' prepended to the file path. Removing changing to './' fixes the problem


koistya:
@abramz thanks!

abramz:
:smile_cat:

PR #90, opened 2015-04-08
Update BrowserSync logging prefix to `RSK`

shakyShane:
In the same way that Google's Web Starter Kit does, you can easily remove the unfortunate `BS` acronym from the BrowserSync CLI output.

My PR proposes that you change the `logPrefix` to `RSK` as seen below:

RSK


koistya:
@shakyShane cool, thanks!

Issue #56, opened 2015-01-24
Split single Gulpfile into per task files


sdiaz:
Hello,

Im using this starter kit for our projects, but we have to modify the current gulpfile and split into files per task in a gulp/ directory for better understanding of build processes as we have modify some of them (we use modernizr, semantic-ui, sass ..)

How do you feel about switching to that model instead of a single file.

The structure would be

gulpfile.js
gulp/index.js
gulp/config.js
gulp/tasks/<tasks>
gulp/util/errorHandling.js

sdiaz:
The gulp boilerplate of the project has been significantly reduced so I don't know if this is still useful and submit a PR or we should keep at is now.

koistya:
FYI, `styles` and `autoprefixer` stuff is going to be removed from the `gulpfile.js` (see #43), leaving just a few tasks related to firing up webpack and development web server. The `gulpfile.js` will be even smaller than it is now. Splitting it into separate files will add unnecessary complexity to the project's structure. Some folks go even farther and remove dependency on Gulp in favor of just using [webpack](https://webpack.github.io) and [npm scripts](https://docs.npmjs.com/misc/scripts).

Issue #43, opened 2014-12-15
Remove Bootstrap CSS


koistya:
What do you think about removing Bootstrap CSS from this template? In favor of some custom minimum basic styles.

ChaosD:
i toyed with http://purecss.io/ for some time.
@koistya, are you planning to turn this in a yo generator? Otherwise, beside from being a lot of additional work for you anyways, supporting multiple frameworks might get quite messy.

yemi:
The less opinionated the better. I would go for custom minimum basic styles since people most likely will use their css lib/framework/style of choice anyways.

codedsignal:
The more dependencies on 3rd party libraries you include the harder it is to understand the structure and intent of the code. +1 on getting rid of Bootstrap in favour of a minimal custom stylesheet.

koistya:
**Spoiler**: The CSS will follow this naming convention:

```css
.ComponentName { }
.ComponentName--modifier { }
.ComponentName-elementName { }
.ComponentName-elementName--modifier { }
```

Which is well suited for modular architecture. No shared styles or images whatsoever. Only styles per component. Btw, the same CSS naming convention is currently used at Twitter.

codedsignal:
@koistya going BEM strikes me as a good move and completely appropriate to React's strength in creating modular components

sdiaz:
It would be great, and what about BEM CSSO tool?

koistya:
@sdiaz, I'm not sure about [CSSO](https://bem.info/tools/optimizers/csso/), I will check it. As of now, the starter kit is using [csscomb](http://csscomb.com/) and [clean-css](https://github.com/jakubpawlowicz/clean-css). Does CSSO have any advantages over these two?

radum:
I agree with @yemi just remove it and people will use whatever they want.

seemsindie:
I agree that it's should be  less opinionated, let the people use whatever they want.

goatslacker:
Bootstrap is good enough to provide as a base. FWIW this should all be easily pluggable.

ivanoats:
:+1: on making it easy to include other CSS frameworks like node-bourbon https://www.npmjs.com/package/node-bourbon

joybro:
-1 for the following reasons.

1. It is very useful for "starters" like me to have bootstrap as default.
2. It would be not that hard for "experienced people" to remove or replace bootstrap.

koistya:
[...]

PR #95, opened 2015-04-15
separate gulpfile

tadjik1:
Hi there.
This pr trying to solve #56 

1. Separate gulpfile and create config file for common configs (eg autoprefixer).
2. Change eslintrc for tests (right now there are many errors about missin 'use strict').


koistya:
@tadjik1 thanks for the PR! FYI, `styles` and `autoprefixer` stuff is going to be removed from the `gulpfile.js` (see #43), leaving just a few tasks related to firing up webpack, development web server. The `gulpfile.js` will be even smaller than it is now. Splitting it into separate files will add unnecessary complexity to the project's structure. Some folks go even farther and remove dependency on Gulp in favor of using just [webpack](https://webpack.github.io) and [npm scripts](https://docs.npmjs.com/misc/scripts).

tadjik1:
Heh)
Actually I believe that gulp allows to do more complex thing with less pain than webpack. For example, sometimes you need to run some utilities task (like generating sitemap, or create zip archive with your project files, or deploy). As for me - I prefer using gulp for that.

koistya:
Gulp doesn't play way with component-based UI development and generally should not be used for bundling your source code (unless you have a flat site structure without components). See:

- http://webpack.github.io/docs/stylesheets.html
- https://github.com/webpack/css-loader
- https://github.com/webpack/style-loader
- https://github.com/webpack/file-loader
- https://github.com/webpack/url-loader

tadjik1:
ok

PR #96, opened 2015-04-22
Changed gulp implementation

mgmcdermott:
Hey everyone,

Newer to React but I've been using this code to familiarize myself and decided to make a PR for the implementation of gulp that I personally use (in case everyone prefers it). It's not an original idea, I know of a few other projects that use this format.

Anyways, as it says in the commit, I separated all gulp tasks into gulp/tasks/. The gulpfile.js now just requires gulp/index.js which then iterates the tasks directory for all .js files and requires them. All shared variables that were in gulpfile.js are now in gulp/config.js. I have tested gulp and all tasks still work as expected. I personally use it to make tasks more readable and easier to add.


mgmcdermott:
Just noticed [your comment on #56](www.github.com/kriasoft/react-starter-kit/issues/56#issuecomment-93260943)

Issue #110, opened 2015-05-15
Have you considered writing CSS in JS? (helps to remove the folder for each component)


totty90:
Hy!
Usually I write encapsulated components with CSS in JS. Take a look at this lib: https://github.com/hackhat/smart-css You are doing basically almost the same thing where you put the css file near the js file, therefore you could put the CSS into the JS. This could remove the need to have a folder for each component (exception if you need not-shared assets)

Also your long css names could be more beautiful `.ComponentName-elementName--modifier { }` (<- that's quite ugly IMO). Take a look at this [example](https://github.com/hackhat/tic-tac-toe-flux/blob/dev/src/js/components/Tile.jsx). I understand that with LESS, this is easier to do, but please consider looking into this approach too because I found it a delight to work with.

I really appreciate the help you are giving here with this lib!

koistya:
You, you're right, long CSS class names are ugly. This has been fixed with the introduction of CSS Modules. What do you think about it? Inline CSS looks good, not sure how practical it is since plain CSS is better supported by text editors and IDEs.

I'm going to close this one as it's not a real issue or feature request. But feel free to continue the conversation on this thread. You're also welcome to participate in discussions on [Gitter](https://gitter.im/kriasoft/react-starter-kit), [Appear.in](https://appear.in/react), or [StackOverflow](https://stackoverflow.com/tags/react-starter-kit).

PR #112, opened 2015-05-17
Pre-render critical CSS

koistya:
- [x] Load application bundle (`app.js`) after HTML/CSS content is loaded
- [x] Pre-render critical CSS to avoid [FOUC]
(http://en.wikipedia.org/wiki/Flash_of_unstyled_content)
- [x] Extract only CSS which is actually used on any particular page
- [x] Add a method for rendering dynamic CSS
- [x] Fix Not Found page not working
- [x] Update npm modules

### Sample Component

```js
// MyComponent.js
import styles from './MyComponent.less'; // Static CSS
import { withCss } from '../decorators';

@withStyles(styles)
class MyComponent {
  render() {
    // Render dynamic CSS
    let color = '#'+Math.floor(Math.random()*16777215).toString(16);
    this.renderCss(`.MyComponent {background: ${color}}`);

    return (
      <div className="MyComponent">
        <div className="MyComponent-foo">Hello</div>
      </div>
    );
  }
}

export default MyComponent;
```

```less
// MyComponent.less
.MyComponent {
  border: 1px solid red;
  &-foo {
    color: green;
  }
}
```

**Note**: `'./MyComponent.less'` is loaded with `css-loader` inside a client-side bundle, and with `style-loader` inside a server-side bundle.

### Server-side Rendering Example

```js
server.get('*', (req, res) => {
  let title;
  let css = [];
  let html = React.renderToString(<App
    onSetTitle={val => title = val; }
    onInsertCss={val => css.push(val); } />);
  res.send(`
    <html>
      <head>
        <title>${title}</title>
        <style>${css.join('')}</style>
      <head>
      <body>
        <div id="app">${html}</div>
        <script src="/app.js"></script>
      </body>
    </body>`);
});
```
Ref #4, #110


src/app.js
```diff
@@ -34,7 +34,11 @@ function run() {
     onPageNotFound: emptyFunction
   };
   let element = React.createElement(App, props);
-  React.render(element, document.getElementById('app'));
+  React.render(element, document.getElementById('app'), () => {
+    console.log('removing #critical-css...');
+    let css = document.getElementById('critical-css');
+    document.getElementsByTagName('head')[0].removeChild(css);
```
Frenzzy: This can simplify:
```js
css.parentNode.removeChild(css);
```
ref: https://developer.mozilla.org/en/docs/Web/API/Node/removeChild

PR #120, opened 2015-05-25
Update webpack.config.js

g00fy-:
this significantly reduces `app.js` size

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -111,6 +111,9 @@ var appConfig = _.merge({}, config, {
   output: {
     filename: 'app.js'
   },
+  module: {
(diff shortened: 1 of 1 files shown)


korczis:
:+1: 

Just tried this and it works like a charm! Nothing got broken. :-)

constantx:
using `gulp build --release` seems to produce an incorrect `app.js` file :(

```
Asset     Size  Chunks             Chunk Names
    app.js  2.29 kB       0  [emitted]  main
```

koistya:
In my understanding this setting is only supposed to be used with large libraries which you're willing to reference directly on a web page (as opposed to bundling it into the `app.js` file).

PR #122, opened 2015-05-27
Fixes gulp race condition

seriallos:
I ran into a race condition when trimming down the provided application.  If the frontend
bundle finished far enough ahead of the server, the server would fail to start (as that
bundle had not finished processing).

This PR waits for all configs to be bundled before moving onto the next step and solves
this race condition.

Minimal application that triggered the error is in my fork of the repo:
https://github.com/seriallos/react-starter-kit-minimal

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -56,11 +56,14 @@ gulp.task('bundle', function(cb) {
   var config = require('./webpack.config.js');
   var bundler = webpack(config);
   var verbose = !!argv.verbose;
+  var numBundled = 0;
 
   function bundle(err, stats) {
     if (err) {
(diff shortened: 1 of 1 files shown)


koistya:
Good catch! With this approach the `started` var becomes obsolete, right?

seriallos:
I'm not 100% sure if it's obsolete or not.  Right now, changes to files while watching re-trigger `bundle` but won't execute the callback.  I haven't played enough with the gulp process to know if that callback has any useful/detrimental side effects after initial startup.

koistya:
@seriallos thanks! I have merged it with a minor fix.

PR #123, opened 2015-05-27
ESLint/Jest Integration Fixes

seriallos:
By default, ESLint fails jest tests because of the global variables available.

This PR adds an extended configuration that is used when running tests.  Normal linting ignores tests to avoid not flagging an undeclared variable in regular JS.


signalwerk:
nice worked for me

.eslintrc.jest
```diff
@@ -0,0 +1,19 @@
...
+    "config": true,
+
+    "afterEach": true,
+    "beforeEach": true,
+    "describe": true,
+    "it": true,
+    "pit": true,
+    "require": true,
+    "xdescribe": true,
+    "xit": true
```
tayzlor: Think you might also need to add here - 

```
    "runs": true,
    "waitsFor": true,
```

koistya:
There is a pre-configured "jest" [environment settings](http://eslint.org/docs/user-guide/configuring#specifying-environments) in ESLint, plus there is a new feature coming in ESLint allowing to have [glob base configuration](https://github.com/eslint/eslint/issues/3611).

Issue #116, opened 2015-05-20
How to implement routing and navigation


koistya:
##### [docs/recipes/how-to-implement-routing.md](https://github.com/kriasoft/react-starter-kit/blob/master/docs/recipes/how-to-implement-routing.md)
- [x] Basic routing
- [x] Asynchronous routes
- [x] Parametrized routes
- [x] Setting page title and meta tags
- [ ] Redirects
- [ ] Code splitting
- [ ] Nested routes
- [ ] Integration with Flux
- [x] Server-side rendering

See also [Universal Router](http://slides.com/koistya/universal-router/fullscreen)
##

Sunil6591:
Why can just use ```react-router```

9mm:
Yeah I was wondering the same thing, how do I use `react-router`? This isn't vert documented as is a bit challenging to figure out.

MattReimer:
I'm trying to figure out how to redirect based on an action. 

Most of the posts Ive seen about `react-router` suggest getting the router from the `this.context` object but mine is empty. Any ideas?

subodh-malgonde:
@koistya Any idea when the docs would be updated? Especially the docs for nested routes and code splitting.

karlstanton:
+1

zomars:
I agree with the `react-router` approach.

FakhruddinAbdi:
react-router  +1

zazapeta:
react-router +1

zazapeta:
Nice demonstration :+1:

rnjailamba:
@koistya  please update docs with Redirects example.

yookore:
What do we need to do to encourage the completion of these docs?

langpavel:
I'm sure something of goals are met. Other are described precisely in standalone issues. Can everybody interesting into this update, reference other issues?

langpavel:
@koistya I think more or less all goals mentioned here are achieved, except redirects.

Code splitting is demonstrated in `feature/react-intl`, not in router, but technically it is same..

ulani:
@koistya  thank you very much for crating this issue! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

NOTE: The `main` branch has been updated with React Starter Kit v2, using JAM-style architecture.

PR #137, opened 2015-06-13
Add how to implement routing recipe

koistya:
- Basic routing
- Asynchronous routes
- Parametrized routes

See [/docs/recipes/how-to-implement-routing.md](https://github.com/koistya/react-starter-kit/blob/routing-doc/docs/recipes/how-to-implement-routing.md), ref #116

### DIFF
--- a/docs/recipes/how-to-implement-routing.md
+++ b/docs/recipes/how-to-implement-routing.md
@@ -1,0 +1,139 @@
+## How to Implement Routing in React.js Apps [![img](https://img.shields.io/badge/discussion-join-green.svg?style=flat-square)](https://github.com/kriasoft/react-starter-kit/issues/116)
+
+ * [Step 1: Basic Routing](#step-1-basic-routing)
+ * [Step 2: Asynchronous Routes](#step-2-asynchronous-routes)
+ * [Step 3: Parametrized Routes](#step-3-parametrized-routes)
+ * Step 4: Handling Redirects
+ * Step 5: Setting Page Title and Meta Tags
+ * Step 6: Code Splitting
+ * Step 7: Nested Routes
+ * Step 8: Integration with Flux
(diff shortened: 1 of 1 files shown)


boda2004:
Common use case for router usage according to https://github.com/koistya/react-starter-kit/blob/routing-doc/docs/recipes/how-to-implement-routing.md will be
```javascript
  try {
    const path = window.location.hash.substr(1) || '/';
    const component = await router.match(path) || <NotFoundPage />;
    React.render(component, container);
  } catch (err) {
    React.render(<ErrorPage {...err} />, container);
  }
```

What if router covers this case by default? 
So code above would look like
```javascript
const path = window.location.hash.substr(1) || '/';
router.match(path);
```
This approach will save some typing and will provide transparent to client code redirects support.

For edge cases match will accept optional second parameter callback, which allows custom requests handling.

This approach will require to extend router initialization and provide additional data:
- root element
- "not found" component
- error component (in case of exception)

grigory-leonenko:
What about clean boilerplate about location has catching?
```javascript
router.run(router.history, async (path) => {
    const matched = await router.match(path);
    React.render(matched, container);
}) 

```
How about render errors and try, catch? Error need be thrown out.

```javascript
try {
    const path = window.location.hash.substr(1) || '/';
    const component = await router.match(path) || <NotFoundPage />;
    React.render(component, container);
  } catch (err) {
    throw new Error(err); 
    React.render(<ErrorPage {...err} />, container);
  }
``` 
If i want to use Router with Flux, i need to get all the parameters of route matching from one place.
```javascript
router.run(router.history, async (route) => {
     let {path} = route;
    const matched = await router.match(path);
    RouterActions.init(route);
    React.render(matched, container);
}) 
```
For example i need to check if user is auth and don't do render before get it. Something like global interceptor with opportunities to redirect.

PR #141, opened 2015-06-15
Removed unnecessary page load at client boot

d0ntth1nc:


koistya:
@d0ntth1nc but.. in order to generate a virtual DOM tree, it needs to has access to data from stores.

d0ntth1nc:
I think that we can pass the initial data in script tag with type "application/json" and parse it on the client side without new request, what do you think?

koistya:
This way you won't be ably to enable strict content security policy (CSP)
on your site, right?
>
>

d0ntth1nc:
using json.stringify + json.parse ofc

d0ntth1nc:
im not sure, imma try some stuff

d0ntth1nc:
Здравствуй! I found nice solution... CSP prevents inline scripts, but not hidden divs. JSON is just a string and we can insert it in <div id="initial-data" style="display:none;"></div> tag. On the client side we can do -> var data = JSON.parse(document.getElementById("initial-data").innerHTML);

koistya:
This approach can potentially has an issue with SEO...

d0ntth1nc:
I found a library named "iso". It creates empty div tag and puts all the data in "data-state" attribute. I don`t think that empty div tag will affect SEO. Neither CSP will prevent empty script tag with data-state attribute. :)

koistya:
This could work. I personally use this approach in my projects - all bootstrap / state data is requested simultaneously with the main application bundle:

```html
<script src="/data.js"></script>
<script src="/app.js"></script>
```

PR #143, opened 2015-06-18
Change var to let

jartek:
`let` is used everywhere else in the file except in this one place


koistya:
@jartek shouldn't it be `const`?

PR #151, opened 2015-06-26
Update Database.js

vachix:
Fix don't call resolve the second time if an error was occoured


koistya:
@vachix thanks!

mychaelgo:
after merge with this commit, i can't access directly to page what i want

Issue #152, opened 2015-06-27
TypeError: Cannot read property 'path' of null


polytechnic:
New clone, When i open new blank window and load component pages contact and login from

http://localhost:3000/contact
http://localhost:3000/login

 i get returned:

```
TypeError: Cannot read property 'path' of null
   at Object.ID_1 (/Users/username/Sites/myapp/build/webpack:/src/stores/AppStore.js:69:26)
   at Dispatcher.$Dispatcher_invokeCallback (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:220:35)
   at Dispatcher.dispatch (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:195:14)
   at /Users/username/Sites/myapp/build/webpack:/src/core/Database.js:45:18
   at run (/Users/username/Sites/myapp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:89:1)
   at /Users/username/Sites/myapp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:100:1
   at process._tickCallback (node.js:355:11)
```

And on the terminal
```
TypeError: Cannot read property 'path' of null
    at Object.ID_1 (/Users/username/Sites/myapp/build/webpack:/src/stores/AppStore.js:69:26)
    at Dispatcher.$Dispatcher_invokeCallback (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:220:35)
    at Dispatcher.dispatch (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:195:14)
    at /Users/username/Sites/myapp/build/webpack:/src/core/Database.js:45:18
    at run (/Users/username/Sites/myapp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:89:1)
    at /Users/username/Sites/myapp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:100:1
    at process._tickCallback (node.js:355:11)
TypeError: Cannot read property 'path' of null
    at Object.ID_1 (/Users/username/Sites/myapp/build/webpack:/src/stores/AppStore.js:69:26)
    at Dispatcher.$Dispatcher_invokeCallback (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:220:35)
    at Dispatcher.dispatch (/Users/username/Sites/myapp/node_modules/flux/lib/Dispatcher.js:195:14)
[...]

Issue #159, opened 2015-07-01
Route Path Error


mzparulina:
Clicked Login then ask question link near footer, clicked the back button then i got error

TypeError: Cannot read property 'path' of null
   at Object.ID_1 (/Users/poline/git/MyApp/build/webpack:/src/stores/AppStore.js:69:26)
   at Dispatcher.$Dispatcher_invokeCallback (/Users/poline/git/MyApp/node_modules/flux/lib/Dispatcher.js:220:35)
   at Dispatcher.dispatch (/Users/poline/git/MyApp/node_modules/flux/lib/Dispatcher.js:195:14)
   at /Users/poline/git/MyApp/build/webpack:/src/core/Database.js:45:18
   at run (/Users/poline/git/MyApp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:89:1)
   at /Users/poline/git/MyApp/build/webpack:/~/babel-core/~/core-js/modules/es6.promise.js:100:1
   at process._tickCallback (node.js:419:13)
![screen shot 2015-07-01 at 11 36 59](https://cloud.githubusercontent.com/assets/5786822/8447062/8744264a-1fe5-11e5-8957-9e4fdac8adb0.png)
##

mychaelgo:
@mzparulina  change code to this https://github.com/kriasoft/react-starter-kit/pull/151

ulani:
@mzparulina   thank you very much for crating this issue! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

NOTE: The `main` branch has been updated with React Starter Kit v2, using JAM-style architecture.

PR #165, opened 2015-07-05
fix issues #152 and #159

deslee:
This seems to resolve issues #152 and #159 while removing unnecessary calls to the server for pages without content (Contact, Login, Register, NotFound)

1. Added a DefaultComponents file, which provides a dictionary indicating pages that do not require a HTTP request
2. Updated AppActions.js (the action creator) to prevent making a GET request if the path exists in the DefaultComponents dictionary
3. Updated AppStore.js, the pages dictionary is now initialized to contain the paths that do not require a HTTP request. These paths now map to a default page object for App.js to use without throwing an error.
4. Updated App.js, replaced the switch statement with a call to the AppStore
5. Refactored Database.js and server.js
    * I borrowed the code from vachix's [pull request 161](https://github.com/kriasoft/react-starter-kit/pull/161)

### DIFF
--- a/src/core/Database.js
+++ b/src/core/Database.js
@@ -6,6 +6,7 @@ import jade from 'jade';
 import fm from 'front-matter';
 import Dispatcher from './Dispatcher';
 import ActionTypes from '../constants/ActionTypes';
+import DefaultComponents from '../content/DefaultComponents';
 
 // A folder with Jade/Markdown/HTML content pages
 const CONTENT_DIR = path.join(__dirname, './content');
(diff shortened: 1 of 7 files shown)


src/core/Database.js
```diff
@@ -29,25 +30,31 @@ export default {
...
+          } else {
+            reject(err);
+          }
+        });
+      }
     }).then((page) => {
       Dispatcher.dispatch({
         type: ActionTypes.RECEIVE_PAGE,
-        page: page});
+        page: page}
```
constantx: strange `}`

constantx:
after this PR, where's a good place to implement routing, as in #137?

koistya:
Thanks for the PR! But with the latest updates it's diverged too much from the `master` branch making it harder to merge. For example, the file which you named `DefaultComponents.js` is called `routes.js` in `master`; `AppActions.js`, `AppStore.js`, `Database.js` files were removed in order to reduce overload in the starter kit.

PR #168, opened 2015-07-10
Fix for gulpfile and webpack config

tomaszkepa:
- Fix gulpfile.babel.js no more error: "TypeError: undefined is not a function" ;)
- Add compile script for gulpfile.js & webpack.config.js
- Updated dependencies and add script for babel in package.json


koistya:
@tomaszkepa thanks for the PR.. but why do you want to pre-compile gulpfile.js? Gulp starting from 3.9+ can read ES6 code just fine..

tomaszkepa:
@koistya ok my mistake ;) because I have gulp version < 3.9 and the project doesn't work for me on older version.

Issue #130, opened 2015-06-02
PostCSS


cmnstmntmn:
PostCSS is used by Google, Twitter, Alibaba, and Shopify. Its plugin, Autoprefixer, is one of the most popular CSS processors.

PostCSS can do the same work as preprocessors like Sass, Less, and Stylus. But PostCSS is modular, 3-30 times faster, and much more powerful.

https://github.com/postcss/postcss

koistya:
Just have plugged PostCSS into webpack's bundling pipeline.

cmnstmntmn:
:+1: great

Issue #126, opened 2015-05-29
why less?


cmnstmntmn:
https://css-tricks.com/sass-vs-less/

i switched to sass, buy i'm just courious why less was choosen as a default option since, many frameworks (compass, burbon, neat) are build on sass

tommmyy:
+1

chiefGui:
+1

Aluxian:
+1

eikeco:
PostCSS +1

frob:
Shouldn't this be closed?

langpavel:
Yes, this should be closed

PR #113, opened 2015-05-17
Switch Jade/Lodash to Handlebars, Less to SCSS

joaquingatica:
May be useful for you to create a branch with this libraries instead of the ones you use.

I prefer handlebars over Jade, and SCSS over LESS, and I think a lot of people do.

Hope this helps.

Cheers!

PR #182, opened 2015-07-24
Remove LESS in favor of vanilla CSS via PostCSS/cssnext

koistya:
> PostCSS can do the same work as preprocessors like Sass, Less, and Stylus. But PostCSS is modular, 3-30 times faster, and much more powerful.

- [x] Convert `.less` files to `.css` (CSS4)
- [x] Configure [PostCSS](https://github.com/postcss/postcss) + [cssnext](http://cssnext.io/)
- [x] Remove `less`, `less-loader`

Ref #130, #126, #113


src/components/Footer/Footer.css
```diff
@@ -0,0 +1,44 @@
...
+  max-width: var(--max-content-width);
+  text-align: center;
+}
+
+.Footer-text {
+  color: rgba(255, 255, 255, .5);
+}
+
+.Footer-text--muted {
+  color: rgba(255, 255, 255, .3);
```
MoOx: `color(#fff a(.3))` can be more clear

webpack.config.js
```diff
@@ -95,7 +90,11 @@ const config = {
     ]
   },
 
-  postcss: [autoprefixer(AUTOPREFIXER_BROWSERS)]
+  postcss: [
+    require('postcss-nested')(),
+    require('cssnext')(),
```
MoOx: you should use postcss-nested after cssnext (which handle the @import). Btw, cssnext have a "plugins" option to add you own plugin (the appropriate "cssnext" way)

webpack.config.js
```diff
@@ -95,7 +90,11 @@ const config = {
...
+    require('autoprefixer-core')(AUTOPREFIXER_BROWSERS)
```
MoOx: cssnext is already included in autoprefixer (and cssnext also have a `browsers` option).

webpack.config.js
```diff
@@ -65,10 +64,6 @@ const config = {
         loader: `${STYLE_LOADER}!${CSS_LOADER}!postcss-loader`
```
MoOx: you can use cssnext-loader + cssnext plugins options for nested. (also follow this https://github.com/cssnext/cssnext/issues/80)

maxbates:
Tips for integrating Sass instead? I'm confused by the immediate adoption of PostCSS - it works great + fast, but very little IDE support.

koistya:
@maxbates you would need to install `sass-loader`, then register to handle `.sass` files in `webpack.config.js`, then you can load them with `import './MyComponent.sass` or `import styles from './MyComponent'; @withStyles(styles) class MyComponent { render () { ... } }`

koistya:
[Vote for PostCSS support in WebStorm](https://youtrack.jetbrains.com/issue/WEB-16601)

maxbates:
Great, thanks a lot

webpack.config.js
tobiasmuehl: You meant autoprefixer is included in cssnext I guess.

MoOx: yeah haha sorry :)

iDVB:
Not sure if I missed something, but IS RSK currently using [cssnext](http://cssnext.io/features/)? Cause I see that it seems to be using alot of the same plugins.... but not directly using cssnext.

Is there a reason for that? Seems that cssnext was something that was originally looked into?

PR #185, opened 2015-08-01
Fix React JS Includes 

nobesnickr:


koistya:
@nobesnickr thanks! Merged with fixes (`.idea` usually goes to the [global .gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore))

PR #192, opened 2015-08-06
Update package.json

sreym:
When `gulp build` started error had happened (Ubuntu 14.04.2 x32, node 0.12.7).
Latest version of cssnext solve this problem.


koistya:
:+1:

PR #197, opened 2015-08-09
package.json vs index.js in components' folders

nodegin:
[Review on Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/197)

### DIFF
--- a/src/components/TextBox/package.json
+++ b/src/components/TextBox/package.json
@@ -1,6 +1,0 @@
-{
-  "name": "TextBox",
-  "version": "0.0.0",
-  "private": true,
-  "main": "./TextBox.js"
-}


koistya:
Thanks of the PR! With this approach if you have multiple files opened in your text editor or IDE, all those tabs will be named identically [index.js] making it harder to navigate between opened files. Do you think this won't be a problem for you?

nodegin:
@koistya I think major editor will able to display path name, and possibly will not code many files at the same time, however that might be a problem.

koistya:
@nodegin so, what approach are you settled on? With `package.json` files or `index.js` files?

nodegin:
@koistya I'm using index.js approach since I hate many extra files

everdimension:
To me, having to work with lots of identically named files is a lot worse than having extra files in the project folder.

I guess another solution would be to just remove `package.json` files and call your components like this

```javascript
import App from 'components/App/App'
```
although that might be too verbose.

nodegin:
I think @everdimension solution is acceptable too, better than mine's.

snez:
Definitely agree with @koistya on finding the right file

nathanmarks:
100% with @koistya here, one of my favourite parts if this starter kit is the fact that files are explicit in their naming rather than "index.js" for every tab in my editor.

zhiyelee:
hmm, i found the reason why there are lots of package.json instead of index.js.  Sounds good.  
But  personally I prefer index.js which is simple for me.  Thanks for answer the question @koistya

Kepro:
index.js? :D come on and my brain blow up in my IDE...

zhiyelee:
Days later, I realized that `index.js` is bad idea for me...

j:
Throwing my two cents right now.  I'm also not a fan of throwing package.json files in folders that aren't a true npm package.  Also, wasn't a fan of index.js files as components since I'm a tab whore and file names shrink. I ended up putting an index.js file in both the components and containers directory where I exported the components.

i.e:
```js
// containers/index.js

import App from './App/App';
import Home from './Home/Home';
import Page from './Page/Page';

exports.App = App;
exports.Home = Home;
exports.Page = Page;

```

```js
// routes.js

import React from 'react';
import { IndexRoute, Route } from 'react-router';
import { App, Home, Page } from './containers';


export default function routes() {
  return (
    <Route path="/" component={App}>
      <IndexRoute component={Home} />
      <Route path="page/:id" component={Page} />
    </Route>
  );
}

```

everdimension:
How about this:

```
Component
| -- index.js
| -- Component.js
| -- Component.scss
```

Inside `index.js` file:
```javascript
// Component/index.js
export default from './Component.js'
```

I find this small `index.js` file much easier (and much shorter) to write and extend than a `package.json` file. And since you're still working the `Component.js`, there is no problem of identical filename in the text editor.

whynot123:
@everdimension can you post a practical example where you need to extend `index.js` and `package.json` won't fit?
i think current structure is pretty straight forward
also with index.js as component file fuzzy search will break
to skip writing a lot of files i wrote small cli script to jumpstart components you can check it out [here](https://github.com/whynot123/react-starter-kit-cg)

everdimension:
@whynot123 
I believe this pattern is better for the following reasons:

1. You're probably already using it in other parts of your application. If not... why not?) A flux "feature" might look something like this:

        Feature
        | -- index.js
        | -- FeatureStore.js
        | -- FeatureActions.scss
        | -- FeatureApi.scss
    And you can conveniently export `FeatureStore` as default and some feature actions as named exports inside `index.js`:

    ```javascript
    // Feature/index.js
    export default from './FeatureStore';
    export { fetchFeature, updateFeature } from './FeatureActions';
    ```

2. Same for components. You might have a `ComponentHelper.js` inside the `Component` folder which you might want to make available for other modules as a named export.

3. Even if you don't have the above-mentioned cases, I think having `index.js` is better because it's a javascript file. So you're not polluting your folders with a zoo of different extentions.

4. You said:
    > also with index.js as component file fuzzy search will break

    How do you mean? Your main component files are still named as `ComponentName.js`, so you can search for them easily. Having lots of `index.js` files inside folders is no different for the search than having lots of `package.json` files inside folders.

sthzg:
I wanted to add one problem related to the local `package.json` and tooling. My apologies if this is known or has a workaround (that I don't know and haven't found).

In IntelliJ's IDEs, as soon as there is a local `package.json`, the `NpmUsedModulesInstalled` inspection will underline imports as `Module is not listed in package.json dependencies` (see the React import in the screenshot).

bildschirmfoto 2016-05-06 um 10 42 49

@everdimension 's `index.js` with an `export default from './Foobar'` solves that, but trades it against a `Default export is not declared in imported module`.

bildschirmfoto 2016-05-06 um 10 56 09

Based on what I know so far I prefer the latter, since it seems like not following the `export default` from an `index.js` is a bug in the tooling that can/should be fixed, while reporting a missing dependency in a local `package.json` seems like a sane thing to do (in general).

Like I said, it is possible that there is a solution that I haven't found. There is also the possibility to disable these inspections, but I wouldn't consider that as a long-term solution.

koistya:
I'm thinking about switching to `index.js` files as suggested by @everdimension, except for `core/fetch/package.json` (works better with isomorphic components/classes). Please upvote or downvote this comment to show your attitude or concerns.

Kepro:
Please no.... if you working in IDE you only see index.js index.js index.js index.js ....

everdimension:
@Kepro looks like you didn't read what we're talking about carefully.

You are still going to have and work in `ComponentName.js` files.

cfogelberg:
Here are some screenshots of what `test1/index.js` and `test2\index.js` look like in Sublime build 3118 and WebStorm 2016.2.1

sublime

webstorm

With the exception of fetch (which is a different situation), these images make me agree with use of index.js over package.json + ComponentName.js (fwiw, I've changed all my code locally)

Sidebar observation: the index.js file name pattern is used in routes in RSK already, where RouteName.js  defines a "pseudo-component" for presentational purposes but index.js has the actual routing impl.

codetony25:
Just thinking about this myself and here's a question I was asking myself while thinking about this:

1. Is it structured in a way that is easily maintained in the long run as the application scales?

Going off some of the suggestions here.

> @cfogelberg`s suggestion: 

what test1/index.js and test2\index.js look like in Sublime build 3118 and WebStorm 2016.2.1 
by @cfogelberg 

I see @cfogelberg's approach understandable, where either of our text editors or IDE's will help us know which index.js folder it belongs to. However it may be harder to manage in the future for other developers and when trying to search for these files and not all developers use the same text editors. Although I do like the fact that I don't have to create extra files for a component, it just might not be better in the long run as the application grows.

@j's suggestion:

```
import App from './App/App';
import Home from './Home/Home';
import Page from './Page/Page';

exports.App = App;
exports.Home = Home;
exports.Page = Page;
```

With this approach, I like that we don't have to create an `index.js`, but man, typing `/App/App` or `/Home/Home` really bothers me haha, which is why it's understandable that most people create an `index.js` or `package.json`. I would say that this solution would be easier to maintain when the application gets large, and a developer would just have to remember when importing a component to type the name twice to make sure that they are importing the component.

> @everdimension suggestion:

```
Component
| -- index.js
| -- Component.js
| -- Component.scss
```

So this is actually the type of structure I agree with, mainly because i'm thinking in the long term, this will be easier to maintain and easier for other developers to find this file. The only downside is that we are creating an extra file the just exports the component, which I don't even mind, it literally doesn't take that long to create an index.js, and it's the price i'm willing to pay for a scalable maintainable / manageable file structure.

ronak301:
I will go with index.js 
Reason : 
If u have react and react native projects in same repo - u can not have two package.json with same name entry.
for example: 
one package.json in components/share/package.json
{
name: 'Share'
}
another package.json in action/share/package.json
{
name: 'Share'
}
react: will run fine.
react-native : throws error - duplicate entry @providesModule Share
So i would prefer index.js

codetony25:
We could also use this webpack plugin:

https://www.npmjs.com/package/directory-named-webpack-plugin

This plugin makes it possible to use the name of the directory as the name of the entry file and makes it easier to find. So no more index.js or package.json files.

everdimension:
@codetony25 Tying a project's behavior to a webpack feature is a bad idea for many reasons.
Commonjs is a standard module format and we shouldn't override its behavior with webpack even if it is possible and even if it looks like a helpful feature.

langpavel:
Very old, closing

Issue #196, opened 2015-08-08
EPEERINVALID error on npm install


nathaniellee:
I get the following anytime I try to npm install.

```
npm ERR! Darwin 14.1.0
npm ERR! argv "node" "/usr/local/bin/npm" "install"
npm ERR! node v0.12.6
npm ERR! npm  v2.12.1
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package eslint does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer eslint-plugin-react@3.2.1 wants eslint@>=0.8.0 || ~1.0.0-rc-0
npm ERR! peerinvalid Peer eslint-loader@0.14.2 wants eslint@0.21 - 0.24 || ~1.0.0-rc-0
```

LOG

```
0 info it worked if it ends with ok
1 verbose cli [ 'node', '/usr/local/bin/npm', 'install' ]
2 info using npm@2.12.1
3 info using node@v0.12.6
4 verbose readDependencies loading dependencies from /Users/nlee/Projects/hilu-react-fullstack/package.json
5 verbose install where, deps [ '/Users/nlee/Projects/hilu-react-fullstack',
5 verbose install   [ 'babel',
5 verbose install     'classnames',
5 verbose install     'eventemitter3',
5 verbose install     'express',
5 verbose install     'fastclick',
5 verbose install     'fbjs',
5 verbose install     'flux',
5 verbose install     'front-matter',
5 verbose install     'jade',
5 verbose install     'lodash',
5 verbose install     'normalize.css',
5 verbose install     'react',
5 verbose install     'react-dom',
5 verbose install     'react-routing',
5 verbose install     'source-map-support',
5 verbose install     'superagent',
5 verbose install     'autoprefixer-core',
5 verbose install     'babel-eslint',
5 verbose install     'babel-loader',
5 verbose install     'browser-sync',
5 verbose install     'css-loader',
5 verbose install     'csscomb',
5 verbose install     'cssnext',
5 verbose install     'del',
5 verbose install     'eslint',
5 verbose install     'eslint-loader',
5 verbose install     'eslint-plugin-react',
5 verbose install     'git-push',
5 verbose install     'gulp',
5 verbose install     'gulp-changed',
5 verbose install     'gulp-if',
5 verbose install     'gulp-load-plugins',
5 verbose install     'gulp-rename',
5 verbose install     'gulp-size',
5 verbose install     'gulp-util',
[...]

PR #198, opened 2015-08-09
Fixing #196 and adding 'dotenv' support

joshjung:
Installed react-starter-kit today and ran into two issues that I'm thinking should potentially be fixed/included:
- Issue #196 for which there was as suggested solution.
- Was hoping to be able to control the PORT via a .env file so I added support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 build
 node_modules
 npm-debug.log
+.env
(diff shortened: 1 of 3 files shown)


koistya:
@joshjung thanks for the PR! Being able to control the PORT is needed only during development? I assume hosting providers with Node.js support already have this property set, no?

joshjung:
I'm doing my own hosting and server management through Rackspace and need this option. If it is not that needed for the majority of developers, we can definitely skip it. I'm also using KeystoneJS for a few projects and they had it integrated and it seemed like a good idea!

PR #200, opened 2015-08-09
fix travis error

nodegin:


nodegin:
solved by modifying ```eslint-loader": "^1.0.0"```

PR #204, opened 2015-08-12
Fix IE<10 history api (fix exception on missing replaceState & pushState)

artemkaint:
This change is [Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/204)

### DIFF
--- a/src/core/Location.js
+++ b/src/core/Location.js
@@ -9,9 +9,9 @@ export default {
   navigateTo(path, options) {
     if (canUseDOM) {
       if (options && options.replace) {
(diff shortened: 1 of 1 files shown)


koistya:
@artemkaint thanks for the PR. I need some time to verify it.

Bogdaan:
`window.history` in ie<10 is undefined.

Maybe better use:
```
window.location.href = path;
```

langpavel:
Uff. Is this still actual?

artemkaint:
IE<11 isn't supported yet. I suppose the ticket is obsolete

PR #206, opened 2015-08-13
fix problem with global define

tadjik1:
Lodash mutate source object while merge it.
https://lodash.com/docs#merge


koistya:
:+1:

PR #233, opened 2015-08-24
Update the README.md file

chentaixu:
Added explanations to the tools folder and removed the deprecated 'webpack.config.js' which is no longer there.  The deploy.js is not implemented by default so I added an explanation. This change should make the react-starter-kit easier to understand as of the current version.


koistya:
Cool, thanks!

PR #238, opened 2015-08-27
Use global.WATCH instead of global.watch

MyBoon:
global.WATCH is defined in start.js 
```js
global.WATCH = true;
````

global.watch is `undefined`.


koistya:
:+1:

PR #242, opened 2015-09-01
remove unused dependnecy run-sequence

cheapsteak:
not useful since no longer using gulp
not used anywhere else in project


koistya:
:+1:

PR #245, opened 2015-09-08
Fixed parameterized route example

ademuk:
First argument of route callback is express Request object, not the parameter.


koistya:
Thanks!

PR #252, opened 2015-09-11
Update React to v0.14.0 RC1

rodolfo2488:
New version of react is ready.
http://facebook.github.io/react/blog/2015/09/10/react-v0.14-rc1.html

But, there is a warning when you don't extend react classes from the React.Component, so I updated that as well.


koistya:
Thanks!

chentaixu:
What? No mentioning of childContext, still?

PR #269, opened 2015-09-30
Update Link.js

maysale01:
When rendering a Link component, the binding to handleClick doesn't work due to some weirdness with static methods and the arrow function. I've tested and it works as a regular static method definition for both static invocation of Link.handleClick.bind in other components, as well as in the rendering of the Link component.

[Stack Overflow Discussion](http://stackoverflow.com/questions/32825754/how-to-force-react-es6-static-methods-this-to-be-bound-to-lexical-scope)

### DIFF
--- a/src/components/Link/Link.js
+++ b/src/components/Link/Link.js
@@ -44,7 +44,7 @@ class Link extends Component {
         this.props && this.props.state || null,
         this.props && this.props.to || (link.pathname + link.search));
     }
-  };
+  }
 
   render() {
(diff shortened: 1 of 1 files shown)


koistya:
It seems like after this fix `<a onClick={Link.handleClick}>` links stop working throwing `TypeError: Cannot read property 'props' of null`.

maysale01:
It's the same thing as https://github.com/kriasoft/react-starter-kit/pull/275 which is already merged, so we can close.

PR #277, opened 2015-10-09
don't use lambda function for handleClick

9mm:
Otherwise, the function doesn’t have access to this.props if one
chooses to use this method: `<Link to="/about"><span>About</span></Link>`


koistya:
It seems like this fix was already proposed in #275.

martin-jujou-sirca:
@koistya 275 fix actually breaks it, the function doesnt have access to this.props

PR #284, opened 2015-10-15
Removed the special rules around the react-routing router

gnz00:
No longer need the webpack config rules for react-routing as the ES5 transpiled is now correctly up to date.

Signed-off-by: Alexander Mays maysale01@gmail.com


langpavel:
Outdated

PR #286, opened 2015-10-16
feat(webpack): long term caching with webpack

vishwanatharondekar:
Webpack supports [long term caching](http://webpack.github.io/docs/long-term-caching.html). I have already done this in my project. Changes requires some config additions in `tools/config.js` for webpack and mechanism to replace the main chunk generated in html.js . 

**Ugliest part** of this pull request is the replace. It changes file in your source which causes it to show as modified in git which I really dont like but have to lilve with. Any changes suggested to do it better way will be welcome.

### DIFF
--- a/tools/config.js
+++ b/tools/config.js
@@ -111,12 +111,14 @@ const appConfig = merge({}, config, {
   ],
   output: {
     path: path.join(__dirname, '../build/public'),
-    filename: 'app.js',
+    chunkFilename: DEBUG ? undefined : '[name].[chunkhash].js',
+    filename : DEBUG ? 'app.js' : '[name].[hash].js'
   },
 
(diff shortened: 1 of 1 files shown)


vishwanatharondekar:
I later on found out that this is not working as expected. Since server webpack bundle is already completed when function in plugins run, to apply this changes build has to be run twice. A working (ugly) fix would be to replace directly in build/sever.js.

Better fix would be to tap into the webpack build between server and client bundling. Use data from client bundling in server bundling. This requires changes in bundle.js and probably will slow down since webapck build will be run serially. (I am not sure if it runs parallely internally when we give array as config object)

vishwanatharondekar:
Closing this PR since the problem is already solved in https://github.com/kriasoft/react-starter-kit/pull/331

PR #296, opened 2015-10-26
Bugfix: keep port of server

zhiyelee:
`port` attribute is still being used in https://github.com/kriasoft/react-starter-kit/blob/master/src/core/HttpClient.js#L13

Can't start the server now with `npm start`


koistya:
Thanks!

PR #297, opened 2015-10-29
replace match with search since search is enough here

zhiyelee:
replace match with search since search is enough here
> When you want to know whether a pattern is found in a string use search() (similar to the regular expression test() method); for more information (but slower execution) use match() (similar to the regular expression exec() method).
-[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search)


zhiyelee:
hi @koistya any update about this?  I can close this if not accepted

PR #307, opened 2015-11-03
Scoped packages in webpack server externals

AlastairTaft:
Added an @ character to the check for a package name in the webpack server config externals section. This allows me to correctly include my private scoped package `@alastair/mybrand.lib`.


koistya:
Thanks, Alastair!

PR #309, opened 2015-11-04
I've had a go at adding details on configuration for sublime text

mikeljames:
I'm a sublime user so thought I'd share.

let me know of any changes that might be needed.

Thanks,

Mike


mikeljames:
Nice ;)

koistya:
:+1:

PR #311, opened 2015-11-09
Removed redundant spread vars

chentaixu:
In serverConfig and appConfig, it is redundant to spread the values in config (config.module and config.plugins) which will be merged by lodash anyways.


koistya:
Cool, thanks!

PR #315, opened 2015-11-12
Content fixes to jade files

doug-orchard:
Jade files were compiling the first word in each new line as an element ( eg, < Morbi >...< /Morbi > ). 

Removed all unnecessary line breaks to clean the rendered HTML.


langpavel:
Outdated

PR #317, opened 2015-11-13
Updated HttpClient.js to better respond to problems.

frob:
Changed the way that non-successful status codes are handled. This defaults to resolving to null in cases that are specifically handled and changes the specific handling to 500 and 406 only.

This was so 403 and other benign server response codes do not throw full app breaking errors.


zhiyelee:
why only `500` and `406`? How about `401` `400` ?

langpavel:
Outdated

PR #330, opened 2015-11-26
Fix ESLint warnings

Frenzzy:
- Update ESLint version to fix `Unexpected space before function parentheses`
- Add brackets around `await` to it `Missing space before keyword "await"`
- Remove second blank lines to fix `Too many blank lines at the end of file. Max of 1 allowed`

t


koistya:
It's better to use async methods inside Node.js app.

Frenzzy:
Ok, PR updated!

tools/build.js
```diff
@@ -20,4 +20,3 @@ async function build() {
 }
 
 export default build;
-
```
CombatCode: http://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file

CombatCode: Also, I'm reffering to project settings https://github.com/kriasoft/react-starter-kit/blob/master/.editorconfig#L14

Frenzzy: This file has two newlines at the end instead of one newline, see airbnb/javascript/pull/578

PR #331, opened 2015-11-26
Resolve long term caching by using webpack AssetsPlugin

Frenzzy:
See webpack docs for more info: https://webpack.github.io/docs/long-term-caching.html

Related #286


koistya:
:+1:

PR #335, opened 2015-12-07
Update the style guide

koistya:
https://github.com/koistya/react-starter-kit/blob/style-guide/docs/react-style-guide.md


goatslacker:
:+1:

PR #341, opened 2015-12-09
One really nice benefit of interpolated string literals 

materliu:
One really nice benefit of interpolated string literals is they are allowed to split across multiple lines。Take advantage of it .

### DIFF
--- a/src/components/Html/Html.js
+++ b/src/components/Html/Html.js
@@ -20,12 +20,12 @@ class Html extends Component {
 
   trackingCode() {
     return ({ __html:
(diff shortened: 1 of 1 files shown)


frenzzy:
But this is an inline script and it should not contain spaces and line breaks.
For exmaple see: https://developers.google.com/speed/docs/insights/MinifyResources

CombatCode:
@frenzzy :+1: , rendered code is just invalid because of white spaces and new lines

materliu:
@frenzzy thx, so much.

PR #342, opened 2015-12-10
Add new recipe suction how-to-do-basics

merriam:
Added a new recipe section with some basics when getting started with the kit.   It's hard starting with the kit without a 'basics' section.


koistya:
Thanks! Give me some time to review.

langpavel:
Yay, really unhappy to see nice work like this so late

PR #347, opened 2015-12-17
delete the redundant code.

materliu:
I don't think this code is useful, because we don't use `server.get('port')` later.


materliu:
add morgan and compression.

koistya:
Currently, `server.get('port')` is used in HTTP client utility [here](https://github.com/kriasoft/react-starter-kit/blob/master/src/core/HttpClient.js#L13). Though, it can be replaced with a value coming from `config.js/port` if needed.

materliu:
@koistya  sorry, did'n notice that.

PR #354, opened 2015-12-17
Replace server API with webpack's code splitting functionality

deslee:
I don't expect this PR to be accepted, but I think this is worth sharing and getting feedback.

Instead of using the server's api, I am getting the contents of the jade files with an asynchronous `require` method. This uses the routes-loader from the [React Static Boilerplate](https://github.com/koistya/react-static-boilerplate) to generate the async require statements (to ensure that each .jade content lives in its own file)

Instead of loading the in `routes.js` like this,

```
const content = await http.get(`/api/content?path=${state.path}`);
```

content is now loaded like this:

```
var handler = routes[state.path];
var result = await handler(); // this will return the same object as the "content" variable in the old code
```

The routes object contains code generated by `routes-loader` and the generated asynchronous requires.

My next step for my personal project is to ditch the server altogether and generate static HTML documents to be hosted on services like AWS S3 or any HTTP server.


ulani:
@deslee thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #369, opened 2015-12-26
Fix propType warning in release mode

dfrankland:
Using `npm start -- --release` upon loading `localhost:5000` a warning is
created:

```
Warning: Failed propType: Required prop `path` was not specified in `ContentPage`. Check the render method of `StyledComponent`.
Warning: Failed propType: Required prop `content` was not specified in `ContentPage`. Check the render method of `StyledComponent`.
```

The issue and fix is mentioned here:
https://github.com/facebook/react/issues/4494#issuecomment-125068868

### DIFF
--- a/src/components/ContentPage/ContentPage.js
+++ b/src/components/ContentPage/ContentPage.js
@@ -15,8 +15,8 @@ import withStyles from '../../decorators/withStyles';
 class ContentPage extends Component {
 
   static propTypes = {
-    path: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


okendoken:
I received the same error in a development mode (`npm start`) in browser console when tried to access unexisting page e.g. http://localhost:3000/blahblah

image

The fix proposed by @dfrankland has solved the issue, I am getting 404 from the API

image

so +1 for this PR.

PR #374, opened 2015-12-28
Add classname module for proper concatination several classes

nodkz:
Without `cx` following line
```
<nav className={[s.root, this.props.className])>
```
became improper separated via comma in html
```
<nav class="root_e4f,some_lass_d6e">
```

### DIFF
--- a/docs/react-style-guide.md
+++ b/docs/react-style-guide.md
@@ -122,10 +122,11 @@ Navigation.propTypes = { items: PropTypes.array.isRequired };
 import React, { PropTypes } from 'react';
 import s from './Navigation.scss';
 import withStyles from '../../decorators/withStyles';
+import cx from 'classnames';
 
 function Navigation() {
   return (
-    <nav className={[s.root, this.props.className]}>
+    <nav className={cx([s.root, this.props.className])}>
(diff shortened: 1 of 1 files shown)


wub:
Would this work?

```jsx
const cx = (...classes) => classes.join(' ');

<div className={cx(s.root, this.props.className)}>
```

If so, you wouldn't have to pull in a whole dependency. `classnames` would be better suited if you wanted  deduping/conditional classes, etc.

docs/react-style-guide.md
```diff
@@ -122,10 +122,11 @@ Navigation.propTypes = { items: PropTypes.array.isRequired };
...
 import s from './Navigation.scss';
 import withStyles from '../../decorators/withStyles';
+import cx from 'classnames';
 
 function Navigation() {
   return (
-    <nav className={[s.root, this.props.className]}>
+    <nav className={cx([s.root, this.props.className])}>
       <ul className={s.items}>
         <li className={[s.item, s.selected]}>
```
wub: What about this line?

nodkz:
Ohh. How old is this. Close it.

My decision to use string templates.

PR #384, opened 2016-01-06
Improve webpack `postcss-loader`

dfrankland:
Added [`cssnano`](https://github.com/ben-eb/cssnano) ([the package used by `css-loader`](https://github.com/webpack/css-loader#minification)) to replace `css-loader?minimize` because it supports more and is easier to apply different options. Also, it allows for removing all comments including important comments like `/*! blah */`. Finally, it is easier to update it separately.

Added the `transform` option to `postcss-import` to prevent issues with css `//` comments and `precss@1.4.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "postcss": "^5.0.14",
     "postcss-import": "^7.1.3",
     "postcss-loader": "^0.8.0",
-    "postcss-scss": "^0.1.3",
     "precss": "^1.4.0",
     "raw-loader": "^0.5.1",
     "react-transform-catch-errors": "^1.0.0",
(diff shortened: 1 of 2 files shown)


koistya:
This may impact source maps? Have you tried adding `postcss-scss` parser to `precss` as described here https://github.com/jonathantneal/precss#usage ?

dfrankland:
Yes I have added the `postcss-scss` package and it actually doesn't work as intended. Single line comments won't work unless they are at the top of the file using it.

MarkMurphy:
How about merging this in asap or adding a comment about `//` in scss files in the current readme. I just wasted a lot of time trying to figure out what the problem was.

MarkMurphy:
@dfrankland This didn't work for me.

PR #387, opened 2016-01-07
Disable Browsersync UI and Notifications in `--release` mode

dfrankland:
This prevents Browsersync UI  and notifications from showing in production.

### DIFF
--- a/tools/start.js
+++ b/tools/start.js
@@ -17,6 +17,8 @@ import webpackConfig from './webpack.config';
 import clean from './clean';
 import copy from './copy';
 
+const DEBUG = !process.argv.includes('--release');
+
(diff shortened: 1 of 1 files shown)


tools/start.js
```diff
@@ -81,6 +83,8 @@ async function start() {
         if (!err) {
           const bs = Browsersync.create();
           bs.init({
+            notify: DEBUG,
+            ui: DEBUG ? { port: 3001 } : false,
```
koistya: Is there a way to keep it using default UI port when DEBUG = true?

dfrankland: Yes, definitely. I'll update it over the weekend.

PR #411, opened 2016-01-18
Fix global css issue came after moved to css-module

vishalvijay:
We can't build a component for `browserupgrade`, so we have to make it a global css class


koistya:
:+1:

PR #414, opened 2016-01-19
Fixed universal rendering support with material-ui.

puradox:
The Material UI set of components require that `navigator.userAgent` be defined before rendering styles on the server side.

https://github.com/callemall/material-ui/pull/2007

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -44,6 +44,10 @@ server.get('*', async (req, res, next) => {
       onPageNotFound: () => statusCode = 404,
     };
 
+    global.navigator = {
+      userAgent: req.headers['user-agent']
+    };
+
     await Router.dispatch({ path: req.path, query: req.query, context }, (state, component) => {
       data.body = ReactDOM.renderToString(component);
       data.css = css.join('');


bjfletcher:
@puradox Hello! Well, it's really great to see this PR - RSK and Material UI is a great combination (-:

A couple of things:

1. perhaps add a comment above the code to explain its purpose;
2. even with your change, I still get the following in the build log: `Warning: Material-UI: userAgent should be supplied in the muiTheme context for server-side rendering.`

Are you on the latest for RSK and Material UI? I'm on today's `master` for RSK and `0.14.4` for Material UI.

Hopefully we'll get this worked out at some point.

inian:
@bjfletcher I am using a modified version of RSK and the latest version of Material UI for my project and I can confirm that this change disables the warning for me.

koistya:
It will set User Agent globally for all users browsing the site. Maybe it's better to put `global.navigator = '';` at the top of `server.js` or something like that.

puradox:
@koistya I agree, it would be better to just set this on the server side.

bjfletcher:
An alternative PR that works for me to address the same problem: https://github.com/kriasoft/react-starter-kit/pull/453

koistya:
Thanks for the PR! Closing in favor of #453.

scozv:
hi, is it worth to set separated userAgent for muiTheme, see https://github.com/callemall/material-ui/issues/3831 ?

That is we add code below in `server.js`, I use feature/redux branch :

```
import React, { PropTypes } from 'react';
import getMuiTheme from 'material-ui/styles/getMuiTheme';
import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';

app.get('*', async (req, res, next) => {
    const muiTheme = getMuiTheme({userAgent: req.headers['user-agent']});
    ....
    await match(routes, {
      render(component, status = 200) {
        ...
        data.body = ReactDOM.renderToString(<MuiThemeProvider muiTheme={muiTheme}><div>{component}</div></MuiThemeProvider>);
        data.css = css.join('');
        return true;
      },
    });
```

rdewolff:
Still having this issue. Asked directly in https://github.com/kriasoft/react-starter-kit/pull/453

PR #446, opened 2016-02-17
update eslint and eslint loader

inian:


frenzzy:
but `eslint-config-airbnb` does not support eslint v2 yet, let's wait airbnb/javascript/pull/730

inian:
Ah I see..I thought eslint loader was the only dependency on eslint.  
And I guess the tests would have to be re-written too

inian:
Will submit a new PR..getting some messy merge conflicts with mine..

PR #448, opened 2016-02-19
Store configurations for client and server apps separately

koistya:
This should prevent accidentally leaking confidential configuration settings, a fail-safe approach.

##### `config/config.client.js`
```js
export const appName = 'React Starter Kit';
```

##### `config/config.server.js`
```js
export const databaseUrl =
  process.env.DATABASE_URL || 'postgresql://postgres@localhost:5433/database';

export const facebook = {
  id: process.env.FACEBOOK_ID || '<your-facebook-id>',
  secret: process.env.FACEBOOK_SECRET || '<your-facebook-secret>',
};

export { appName } from './config.client.js';
```

##### `config/package.json`
```json
{
  "name": "config",
  "main": "./config.server.js",
  "browser": "./config.client.js"
}
```

Note: The same approach is currently used in the HTTP client utility (`src/core/fetch`).

### How to Use

```js
import { appName, databaseUrl } from '../config';

console.log(appName); -> available in both client-side and server-side code
console.log(databaseUrl); -> available only in server-side code
```

**Pro Tip**: If you're using PostgreSQL, you can set password for the `postgres` user from the connection string above via [.pgpass](http://www.postgresql.org/docs/current/static/libpq-pgpass.html) file, so that the same connection string above could work for all team members. Also, all hosting providers (e.g. Heroku, OpenShift, Azure, Bluemix) allow you to configure environment variables, no need to store any actual passwords in your GitHub repository.

For review @frenzzy, @Bogdaan, all.


Bogdaan:
:white_check_mark: good approach )

koistya:
Mm.. over-engineering! As it is now is good enough.

README.md
```diff
@@ -37,6 +37,7 @@ visit our sponsors:
 │   ├── /actions/               # Action creators that allow to trigger a dispatch to stores
 │   ├── /api/                   # REST API / Relay endpoints
 │   ├── /components/            # React components
+│   ├── /config/                # Configuration options for both client and server apps
```
frenzzy: Did you forget to update this comment?

src/config/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "private": true,
+  "name": "config",
+  "main": "./config.server.js",
+  "browser": "./config.client.js"
```
frenzzy: It would be nice to remove client config if no longer needed

koistya:
Closing in favor of 6878c79dd6974e6d4571601b3a00863b3a850211

PR #449, opened 2016-02-19
Update Google Analytics tracking code

koistya:


frenzzy:
:+1:

davewallace:
@koistya Hi there, where should I be defining the trackingId? I've tried changing `src/config.js`:

`  google: { trackingId: process.env.GOOGLE_TRACKING_ID || 'UA-78694734-1' },`

Also tried:

`  google: { trackingId: 'UA-78694734-1' },`

And also tried editing `src/views/index.jade` (surely this is the incorrect location):

`ga('create','UA-78694734-1','none');ga('send','pageview')`

When I update using config.js the trackingId string in the resulting localhost:3001 page is empty. When I update the jade template it turns up, but in neither case is the GA script tag is never being rendered on the page, I'm guessing because the conditional in the jade template is not met.

PR #450, opened 2016-02-20
Add 'process.env.BROWSER' flag

codeorelse:
Universal JavaScript is pretty awesome, but sometimes you just only want to run code in the browser (e.a. JS animation) or on the server. I a `BROWSER` flag to `'process.env`.

```
// Only run code in browser
if (process.env.BROWSER) {
  ...
}
```

### DIFF
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -120,7 +120,7 @@ const clientConfig = extend(true, {}, config, {
   // http://webpack.github.io/docs/configuration.html#devtool
   devtool: DEBUG ? 'cheap-module-eval-source-map' : false,
   plugins: [
-    new webpack.DefinePlugin(GLOBALS),
+    new webpack.DefinePlugin({ ...GLOBALS, 'process.env.BROWSER': '"true"' }),
     new AssetsPlugin({
       path: path.join(__dirname, '../build'),
       filename: 'assets.js',
@@ -174,7 +174,7 @@ const serverConfig = extend(true, {}, config, {
   },
   devtool: 'source-map',
   plugins: [
-    new webpack.DefinePlugin(GLOBALS),
+    new webpack.DefinePlugin({ ...GLOBALS, 'process.env.BROWSER': '"false"' }),
     new webpack.BannerPlugin('require("source-map-support").install();',
       { raw: true, entryOnly: false }),
   ],


koistya:
In React library itself they use `__DEV__` flag. Maybe it's a good idea to follow the same convention and introduce `__SERVER__` flag..

koistya:
@codeorelse thanks again for the PR, I just looked around how other folks are solving this problem and it seems to me that the best solution would be setting `'production.env.BROWSER': true` via `DefinePlugin` (instead of `__SERVER__` and `__CLIENT__`).

codeorelse:
I prefer `__SERVER__` over `process.env`, because I feel `process.env` is more about the environment the application is running as a whole (e.a. `process.env.LOCAL`). 

I'm not sure we need to add __BROWSER__ explicitly. What do you think?

koistya:
Check this out: http://stackoverflow.com/a/30355080/82686

codeorelse:
Alright! I'll look into it!

koistya:
Another thing, is that instead of using `(typeof window !== 'undefined' && window.document)` it's better to provide a constant (JSON) value. Here is an example:

```
const clientConfig = {
  ...,
  plugins: [
    new webpack.DefinePlugin({
      'process.env': {
        NODE_ENV: DEBUG ? '"development"' : '"production"',
        BROWSER: '"true"',
      },
    }),
  ],
  ...
};
```

```
const serverConfig = {
  ...,
  plugins: [
    new webpack.DefinePlugin({
      'process.env': {
        NODE_ENV: DEBUG ? '"development"' : '"production"',
        BROWSER: '"false"',
      },
    }),
  ],
  ...
};
```

codeorelse:
Hmm, when I add this to the `serverConfig`..
```
const SERVER_GLOBALS = {
  'process.env.BROWSER': '"foo"'
};
....
plugins: [
  new webpack.DefinePlugin(extend(true, GLOBALS, SERVER_GLOBALS)),
  ...
}
```
.. I also get the value `foo` in the browser.

koistya:
`new webpack.DefinePlugin({ ...GLOBALS, 'process.env.BROWSER': '"false"'})`

codeorelse:
Aight, tested it & updated the PR.

If I'd used `Object.assign({}, GLOBALS, {'process.env.BROWSER': '"false"'})`, it would have worked as well. The spread operator I a far more elegant solution, of course :). Thanks for helping me out.

koistya:
Looks good to me! There is a minor glitch - line 111 was removed.

koistya:
@codeorelse cool. ESLint may complain about a missing space char before `})`, also can you [squash](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) your commits into one before I can merge it?

codeorelse:
Thnx!

PR #453, opened 2016-02-21
Tell CSS tooling to use all vendor prefixes if the user agent is unknown

bjfletcher:
This is a [requirement](http://www.material-ui.com/#/get-started/server-rendering) of Material UI otherwise the user may not get properly styled components.

### Steps to reproduce

With the React Starter Kit, I add Material UI as follows:

    npm install material-ui --save

and the following import to a component:

    import RaisedButton from 'material-ui/lib/raised-button';

and:

    <RaisedButton label="Default" />

I get the following error:

> Warning: Material-UI: userAgent should be supplied in the muiTheme context for server-side rendering.

According to Material UI's documentation, it says I'd need to define the `userAgent`.

### Pull request

This pull request ensures that the `userAgent` is set to `all` if it is not already defined, along with a comment explaining what it is for.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -20,6 +20,13 @@ import { port } from './config';
 const server = global.server = express();
 
 //
+// Tell any CSS tooling (such as Material UI) to use all vendor prefixes if the
+// user agent is not known.
+// -----------------------------------------------------------------------------
+global.navigator = global.navigator || {};
+global.navigator.userAgent = global.navigator.userAgent || 'all';
+
+//
 // Register Node.js middleware
 // -----------------------------------------------------------------------------
 server.use(express.static(path.join(__dirname, 'public')));


bjfletcher:
Am open to any suggestions if this PR isn't suitable for any reason.

koistya:
@bjfletcher thanks! Looks good to me.

puradox:
Good work!

bjfletcher:
@koistya & @puradox You're both very welcome. :) Thanks for an excellent starter kit.

rdewolff:
In the steps to reproduce, you do not mention anything about ```<MuiThemeProvider>```. How do you integrate it with the rest of the app?

Am asking because I have tried many different option but still have errors : either Material UI is not rendering properly due to different user-agent or I get the warning message that the ```[...] Instead, figure out why the markup being generated is different on the client or server```.

Thanks for sharing more info!

comerc:
It is bad PR for Material-UI. Please see [this case](https://github.com/callemall/material-ui/blob/aee887137d1b11b086b7132e69b0f13b2b1e5b5d/src/utils/autoprefixer.js#L10-L15).

We need custom code only, if use Material-UI:
```jsx
    const muiTheme = getMuiTheme({}, { userAgent: req.headers['user-agent'] })
    const data = { ...route }
    data.children = ReactDOM.renderToString(
      <App context={context}>
        <MuiThemeProvider muiTheme={muiTheme}>
          {route.component}
        </MuiThemeProvider>
      </App>)
```

PR #455, opened 2016-02-21
Update ESLint and deps to latest

inian:


koistya:
I'm wondering if the new version of ESLint is bug free..

image

PR #466, opened 2016-02-26
Fix test errors: single quotes

okendoken:
When running `npm run test` on a latest version from master tests fail with the following error:
image

This PR fixes the issue


koistya:
:+1:

PR #469, opened 2016-02-28
Add support for markdown and html templates

Jazmon:
Adds support for markdown and html template files alongside with jade. 

The code refers to markdown and html already in the comments, but no actual support yet was available.


koistya:
Cool, thanks!

PR #475, opened 2016-03-01
Create config file to client and server

sanches89:
This change is [Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/475)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -13,12 +13,13 @@ import FastClick from 'fastclick';
 import Router from './routes';
 import Location from './core/Location';
 import { addEventListener, removeEventListener } from './core/DOMUtils';
+import { titlePrefix } from './config';
 
 let cssContainer = document.getElementById('css');
 const appContainer = document.getElementById('app');
 const context = {
   insertCss: styles => styles._insertCss(),
-  onSetTitle: value => (document.title = value),
+  onSetTitle: value => (document.title = `${titlePrefix}${value}`),
   onSetMeta: (name, content) => {
     // Remove and create a new <meta /> tag in order to make it work
     // with bookmarks in Safari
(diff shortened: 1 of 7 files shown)


koistya:
Yeah, I had the same PR idea ([#448](https://github.com/kriasoft/react-starter-kit/pull/448)) some time ago, but not sure if the majority of RSK users actually need it, because having a server-only config may work fine in many use cases (if you need to pass data from this config to the client, you can do that via state/context).

sanchesic89:
I can pass data from this configuration file to the client application using the state/context, only when the application is in the server, because I can't access the configuration file in the browser to create the context object to pass across my components.

Example:
```js
// -------- config.js --------
export const appName = 'React Start Kit';
export const pgConfig = 'confidential configuration settings';

// -------- server.js --------
import { appName, pgConfig } from './config';

const context = { appName, /* others variables */ };
/* Pass context to router state */

// -------- client.js --------
import { appName } from './config'; // <-- leaking confidential configuration settings

const context = { appName, /* others variables */ };
/* Pass context to router state */
```

inian:
+1
For example, I am using Stripe which gives both a Stripe publishable key (to be used in the client side) and Stripe Secret Key (to be used only on the server). 
I put both of these in the config.js file and used the stripe publishable key in one of my components by importing it from the config. However, the entire config file gets imported in the client bundle generated by webpack!
Separating a client and a server config bundle would definitely make things more clear.

ImanMh:
@koistya 
I think server config should be in a separate file from client config. I have added my secret configs to the ```serverConfig.js``` and public configs to the ```appConfig.js``` and passed ```appConfig``` to both client and server. I'm also dynamically loading different configs for different "production" or "development" environments. If you like the Idea I can create a pull request and add it to the starter kit.

langpavel:
Sorry, this is really what we don't want.

At first, it can buzz that config can be easily passed to client, but this is not true, because, well client is can be dynamically configured only from server and this config files are evaluated at build time

At second, everything you can configure is code which is rewritten in firsts commits of new adopter of RSK, so I believe you are not sad if I close this as not really useful. :slightly_frowning_face: 

BTW I really appreciate effort and work what you invest in this, but I must say sorry..

PR #483, opened 2016-03-04
Add tree-shaking, webpack2, native node modules

justicepay:
Facing some issues with content routing.


src/routes/home/home.js
```diff
@@ -9,7 +9,7 @@
 
 import React, { PropTypes } from 'react';
 import withStyles from 'isomorphic-style-loader/lib/withStyles';
-import s from './Home.scss';
+import s from './home.scss';
```
langpavel: Why you change case of file name? Does you rename files? Do you think about case-sensitive file systems?

justicepay: https://github.com/justicepay/react-starter-kit/commit/5e3b4c61d0e22d357cebe052db36eb1ff1eb010c
it was a small fix

langpavel: Ok, thanks for answer, I see..

tools/webpack.config.js
```diff
@@ -56,11 +56,17 @@ const config = {
...
   plugins: [
-    new webpack.optimize.OccurenceOrderPlugin(),
+    new webpack.optimize.OccurrenceOrderPlugin(),
   ],
 
   resolve: {
     extensions: ['', '.webpack.js', '.web.js', '.js', '.jsx', '.json'],
+    modules: [
+      path.resolve('./src'),
+      path.resolve('./tools'),
```
langpavel: Is `tools` necessary? I don't think so

tools/webpack.config.js
```diff
@@ -180,9 +195,11 @@ const serverConfig = extend(true, {}, config, {
   plugins: [
     ...config.plugins,
     new webpack.DefinePlugin({ ...GLOBALS, 'process.env.BROWSER': false }),
-    new webpack.BannerPlugin('require("source-map-support").install();',
-      { raw: true, entryOnly: false }),
+    new webpack.BannerPlugin({
+      raw: true,
+      banner: 'require("source-map-support").install();',
+    }),
```
langpavel: is `entryOnly` unnecesary now?

langpavel:
@justicepay Ok, let's summarize my opinions:

Firstly I don't like heavy renaming of files. Commit you mention in inline comment was not included to this PR so it's unmergeable.

Secondly I don't like rewriting all `imports` to `requires`. I prefer import statements.

But more important is that both changes will introduce so much conflicts in other users code.

Only plus is try to upgrade webpack, but this also needs some notes what will change for end users of RSK.

justicepay:
@langpavel 
1) Ok, my mistake. That day I couldn't wait for this small fix.
2) But the point of this PR was that we should use native module implementations. There's no other options to implement tree shaking without excluding es2015 babel presets which have transform-es2015-modules-commonjs solution for import statements. The only solution is likely to be using upcoming node6 which should have native es2015 import statement.

I'll close this PR. Take it as a guide to upgrading.

tools/webpack.config.js
justicepay: https://github.com/webpack/webpack/commit/b76aa12b380ffba004edae23d35942dd84bc4624
https://github.com/webpack/webpack/commit/0f0ede8a2c1db71b85c924aceb4263bb507f1927

tools/webpack.config.js
justicepay: Indeed. I was thinking it was kinda related webpack.config.js file thats in tools directory. If you look at the extensions array, you could see '.webpack.js', yeah its confusing. Yeah, my bad.

justicepay:
@langpavel If you really want to upgrade to the webpack2 without tree-shaking. There's only two changes related to this repo that should be fixed. Its a renamed OccurrenceOrderPlugin and BannerPlugin changes.

langpavel:
@justicepay Hi, now I understand what *tree shaking* is :-D... but I think it's working with import statements, they are transpiled to require and then consumed by webpack, aren't they? May be I missed something. Why transform-es2015-modules-commonjs breaks things? Can you reference any issue on babel or webpack?

justicepay:
@langpavel Everything in tools section is running on the node, in our case on the babel-node which uses commonjs module system that doesn't have tree shaking implementation. In webpack2 they introduced their own implementation of commonjs module system that has tree shaking. The thing is for some reason there's no option of webpack2 to use different presets for bundling or vice versa babel-node cannot use different presets either.
Here's the [link](http://www.2ality.com/2015/12/webpack-tree-shaking.html) that might help you understand whats is going on here.

langpavel:
@justicepay Oh, I'm confused from one commit, but there should be actually three:

 1. `import` → `require` in `tools/*.js` – *Question:* Does code still needs babel-node or node5 is enought? It should speed up running of these scripts.
 2. Filename case change – this should not be in PR (merging conflicts)
 3. Webpack 2 – It's Ok and :+1: 

I don't understand why you change import to require and now I see it's only in tools, I also have filename case change in my mind so it confuses me. Now I understand :-D

justicepay:
@langpavel babel-node is needed for transpilation for sure. There's async functions and some of es2015(es6) features that is not implemented in node5.

langpavel:
@justicepay sure, I forgot.. Thanks.

langpavel:
@justicepay please, can you help me with #602?

PR #484, opened 2016-03-06
Update to Jest 0.9.0 and add an example with enzyme.

cpojer:
We just made a huge update to Jest and I figured it is a good idea to update Jest in this repo. I added `babel-jest` which is automatically used and there doesn't need to be a custom preprocessor any longer :)

The documentation also received a major overhaul: http://facebook.github.io/jest/

As an example, I added a test using the enzyme framework ( https://github.com/airbnb/enzyme ) which seems like a pretty good way to test React components. I also added a stub for SCSS files, because those don't work in a testing context :)


koistya:
@cpojer thank you very much for this update! BTW, do you know why there is an error "Error: Cannot find module 'estraverse-fb'"? Also, maybe moving /src/stubs/SCSSStub.js to `/test/stubs/SCSS.js` will make sense?

cpojer:
I added estraverse-fb, for some reason `eslint` needs it to work? I also moved the stub. eslint still fails for some components, however:

```
react-starter-kit/src/components/Footer/Footer.js
  15:1  error  Component should be written as a pure function  react/prefer-stateless-function
```

should this lint rule be disabled or the components be rewritten?

koistya:
Cool, thanks! This is a good rule I believe, will need to rewrite the `Footer` component to make it pass. I'm checking, if there is a smooth way to provide different configs for ESLint if the source code and unit tests are located in the same root folder..

cpojer:
Oh, it turns out that `jest` is its own environment in eslint (nice!) so you can just set `jest: true` in the env variable.

koistya:
BTW, maybe it makes sense to set `/* eslint-env jest */` flag inside the unit test files, as opposed to setting this option in global `.eslintrc` file?

cpojer:
I just pushed an update to the eslintrc file that enables the browser and jest envs. I don't think adding this to every unit test would be fun.

koistya:
@cpojer do you think there is a way to make Jest work with test files located in the same folder with source files? E.g. `/components/Nav.js` + `/components/Nav.spec.js` (unit test for it)? Or, is there a reason not to do so?

cpojer:
@koistya this is currently not possible but I add this to Jest in the near future :)

cpojer:
What would you like me to do here to get this merged? Should I rewrite the components as functional components to make eslint happy?

PR #487, opened 2016-03-08
Use ReactDOMServer for server rendering

inian:
ReactDOM.renderToString is deprecated and removed in React 15. 
I also think this is the only change we need to make to move to react 15.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -15,7 +15,7 @@ import bodyParser from 'body-parser';
 import expressJwt from 'express-jwt';
 import expressGraphQL from 'express-graphql';
 import jwt from 'jsonwebtoken';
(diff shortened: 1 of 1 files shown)


Bogdaan:
:-1: You just rename import name ). This kit already uses react-dom-server

koistya:
The current naming convention suggests the same names for client and server npm modules, for example:

* `ReactDOM` on the client uses `react-dom` module, and `react-dom/server` module on the server
* `fetch` on the client uses `whatwg-fetch` module, and `node-fetch` module on the server

Is there a good reason to change that convention?

inian:
Ah my bad. Got thrown off my the variable name - thought we were just using react-dom because of that.

Issue #505, opened 2016-03-15
[AWS][EC2] Error: Cannot find module 'core-js/library/fn/get-iterator


az-ericso:
I am getting this error after deploying the content in the "build" folder to AWS Elastic Beanstalk. 

``` sh
Error: Cannot find module 'core-js/library/fn/get-iterator'
    at Function.Module._resolveFilename (module.js:326:15)
    at Function.Module._load (module.js:277:25)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.module.exports.exports.__esModule (/var/app/current/webpack:/external "core-js/library/fn/get-iterator":1:1)
    at __webpack_require__ (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:19:1)
    at Object.<anonymous> (/var/app/current/~/babel-runtime/core-js/get-iterator.js:1:19)
    at __webpack_require__ (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:19:1)
    at Object.module.exports.exports.__esModule (/var/app/current/server.js:1555:23)
    at __webpack_require__ (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:19:1)
    at Object.<anonymous> (/var/app/current/webpack:/routes.js:11:1)
    at __webpack_require__ (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:19:1)
    at Object.<anonymous> (/var/app/current/webpack:/server.js:22:1)
    at __webpack_require__ (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:19:1)
    at /var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:39:1
    at Object.<anonymous> (/var/app/current/webpack:/webpack/bootstrap 8dc144ed7f4b43610864:39:1)
```

hansiemithun:
Any updates on this error? Even i am also getting the same error

topwebmaster:
Im too

Qboooogle:
I am getting the same error now.What should I do?

szl-926929:
i'm too

langpavel:
Can this still be an issue? I believe you doing something wrong. Did you build your app and install dependencies on server?

jitendrakhedar:
I am facing the same error on starting the appium server. Any update?

smapig:
[...]

PR #506, opened 2016-03-15
[AWS][EC2] Error: Cannot find module 'core-js/library/fn/get-iterator #505

az-ericso:
I moved the "core-js" module to "dependencies" section from "dev-dependencies".  After that I can successfully deployed the build to AWS Elastic Beanstalk.  I think this fix should not be risky.


koistya:
@az-ericso what version of `npm` are you using in EB?

az-ericso:
Node.js version 2.0.8

64bit Amazon Linux 2015.09 v2.0.8 running Node.js

Node.js 4.3.0
Also supports 0.12.10, 0.10.42, 0.8.28

NPM 2.14.7

Issue #491, opened 2016-03-10
Error: Failed to get mock metadata on npm test: Why this happen?


joowh85:
When I try 

`npm test`

always fail with the errors.
What should I do?

My OS is  Windows 10.

```
> eslint src && jest

Using Jest CLI v0.9.0, jasmine2, babel-jest
 FAIL  components\App\__tests__\App-test.js
● Runtime Error
Error: Failed to get mock metadata: C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\core-js\library\modules\_global.js

See: http://facebook.github.io/jest/docs/manual-mocks.html#content
    at Loader._generateMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:355:15)
    at Loader.requireMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:215:45)
    at Loader.requireModuleOrMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:226:19)
    at Object.<anonymous> (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\core-js\library\modules\_shared.js:1:133)
    at Loader._execModule (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:310:17)
    at Loader.requireModule (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:155:14)
    at Loader._generateMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:347:34)
    at Loader.requireMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:215:45)
    at Loader.requireModuleOrMock (C:\Users\sangw\Documents\Work\jimiarts-web\node_modules\jest-cli\src\HasteModuleLoader\HasteModuleLoader.js:226:19)
[...]

PR #511, opened 2016-03-17
fix(test): add core-js to unmockedModulePathPatterns

JeffreyATW:
This seems to solve the issue seen in #491. I don't see any reason why `core-js` should need to be mocked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,8 @@
     },
     "unmockedModulePathPatterns": [
       "react",
-      "enzyme"
+      "enzyme",
+      "core-js"
     ]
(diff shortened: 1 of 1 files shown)


koistya:
Cool, thanks!

everdimension:
@JeffreyATW 
Were you able to run `npm test` successfully after this change? It just throws another error when I try.

JeffreyATW:
@everdimension Yes, I got my stuff to pass, but I also removed all `Link`s from my Footer, so you might have to unmock `"history"` as well.

PR #519, opened 2016-03-20
Update Home.js

que01:
fix bug,while run "npm start" in a new project throw a error.

---

This change is [Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/519)

### DIFF
--- a/src/routes/home/Home.js
+++ b/src/routes/home/Home.js
@@ -17,7 +17,7 @@ function Home({ news }) {
       <div className={s.container}>
         <h1 className={s.title}>React.js News</h1>
         <ul className={s.news}>
-          {news.map((item, index) => (
+          {(news || []).map((item, index) => (
             <li key={index} className={s.newsItem}>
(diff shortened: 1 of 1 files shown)


src/routes/home/Home.js
```diff
@@ -17,7 +17,7 @@ function Home({ news }) {
       <div className={s.container}>
         <h1 className={s.title}>React.js News</h1>
         <ul className={s.news}>
-          {news.map((item, index) => (
+          {(news || []).map((item, index) => (
```
frenzzy: Why not as [default prop value](https://facebook.github.io/react/docs/reusable-components.html#default-prop-values)?
```js
Home.defaultProps = {
  news: [],
};
```

piglovesyou:
Now we have this so the null check is not necessary.
https://github.com/kriasoft/react-starter-kit/blob/140bd51e8e9b0c10148b632f7cfec3db9643f32e/src/routes/home/index.js#L21

PR #525, opened 2016-03-22
clean build/package.json

whynot123:


tools/cleanPackage.js
```diff
@@ -0,0 +1,20 @@
...
+  const jsonPackage = require('../package.json');
+  await Promise.all([
+    delete jsonPackage.devDependencies,
+    delete jsonPackage.babel,
+    delete jsonPackage.jest,
+    delete jsonPackage.stylelint,
+    jsonPackage.scripts = {
+      start: 'node server.js',
+    },
+  ]);
```
koistya: Why is this wrapped into `Promise.all(...)`?

koistya:
@whynot123 nice PR! What do you think about putting this logic into `tools/copy.js` as opposed to creating a separate file/task for it?

```js
/* ... */
import pkg from '../package.json';

async function copy() {
  /* ... */
  delete pkg.devDependencies;
  delete pkg.babel;
  delete pkg.jest;
  delete pkg.stylelint;
  pkg.scripts = { start: 'node server.js' };
  await fs.writeFile('./build/package.json', JSON.stringify(pkg, null, 2));
}

export default copy;
```

frenzzy:
Maybe it is better to take from root `package.json` only what is really need?
```js
/* ... */
import pkg from '../package.json';

async function copy() {
  /* ... */
  await fs.writeFile('./build/package.json', JSON.stringify({
    private: true,
    engines: pkg.engines,
    dependencies: pkg.dependencies,
    scripts: {
      start: 'node server.js',
    },
  }, null, 2));
}

export default copy;
```

koistya:
@whynot123 looks good! If you can [squash you commits into one](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git), that would be awesome!

whynot123:
@koistya updated according to @frenzzy suggestion
sorry, new to git, 71ee588 will do?

koistya:
@whynot123 as of now, this PR contains [11 commits](https://github.com/kriasoft/react-starter-kit/pull/525/commits). If that's the same in your local repo, you can rollback them and keep the changes like this. Then create a new commit, and push with `--force` flag.

```bash
$ git reset --soft HEAD~11
$ git add .
$ git commit -m 'Update tools/copy.js to clean up package.json before copying'
$ git push origin master --force
```

Issue #334, opened 2015-12-05
How to store data?


AliceRossa:
Hello,

I'm trying to learn React tool and I would like to develop a simple website.

I will wish to create a very simple CRUD: create, read, update and delete an article (text only). I managed to do so with Express and Mongoose but I don't know how with react starter kit.

I installed mongoose:

```
npm install --save mongoose
```

And I added in server.js

```
mongoose.connect('mongodb://localhost/react-starter-kit');
var db = mongoose.connection;
db.on('error', function () {
  throw new Error('unable to connect to database');
});
```

I created a "models" folder and I created "articleModel.js"

```
var mongoose = require('mongoose');

var schema = new mongoose.Schema({
  title : String,
  content : String
});

module.exports = articleModel = mongoose.model('articleModel', schema);
```

How to add a new article then display all articles in the database?

I recently discovered React, thank you for your help.
##

koistya:
So far, I have an example with native MongoDb driver:

https://medium.com/@tarkus/how-to-open-database-connections-in-a-node-js-express-app-e14b2de5d1f8

AliceRossa:
Thank you for your reply !

I tried to do something very simple, I created the user.js file in the API folder

```
import { Router } from 'express';
import { ObjectID } from 'mongodb';

const router = new Router();

router.get('/', async (req, res, next) => {
  try {
    const db = req.app.locals.db;
    const user = await db.collection('user').find();

    if (user) {
      res.send(user);
    } else {
      res.sendStatus(404);
    }
  } catch (err) {
    next(err);
  }
});

export default router;
```
In server.js file I added :
```
server.use('/api/users', require('./api/users'));
```

In route.js file I added :
```
import UsersPage from './components/UsersPage';

on('/users', async () => {
    const content = await http.get('/api/users');
    return content && <UsersPage {...content} />;
  });
```
I created the component, UsersPage.js
```
import React, { Component, PropTypes } from 'react';
import s from './UsersPage.scss';
import withStyles from '../../decorators/withStyles';

const title = 'Users';

@withStyles(s)
class UsersPage extends Component {

  static contextTypes = {
    onSetTitle: PropTypes.func.isRequired,
  };

  componentWillMount() {
    this.context.onSetTitle(title);
  }

  render() {
    return (
      <div className={s.root}>
        <div className={s.container}>
          <h1>{title}</h1>
          <div dangerouslySetInnerHTML={{ __html: this.props.content || '' }} />
        </div>
[...]

PR #534, opened 2016-03-27
Introduce Sequelize as ORM.

arkeros:
First of all I want to congratulate @koistya for your excellent work with this project.

Since this project is a "starter kit", it seems reasonable to support more databases such as MySQL or SQLite, and the simplest way to do going to use an ORM. Therefore, I have decided to collaborate and add Sequelize to this project.

Changes I made:
The default database is sqlite to facilitate the development.
Porting Facebook login to use the new database.
A folder 'models' in './src/data' containing the 'User', 'UserLogin', 'UserClaim' and 'UserProfile' models.

This is the first PR I do. Don't be mean to me if I did something wrong :)

---
This change is [Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/534)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "react": "15.0.0-rc.2",
     "react-dom": "15.0.0-rc.2",
     "react-routing": "0.0.7",
+    "sequelize": "^3.20.0",
     "source-map-support": "0.4.0",
     "whatwg-fetch": "0.11.0"
   },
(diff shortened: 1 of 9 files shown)


arkeros:
This solves https://github.com/kriasoft/react-starter-kit/issues/334#issuecomment-195697330

koistya:
@arkeros good stuff, much appreciated!

5punk:
Nice work, thanks a lot for this.
By any chance could you add some documentation here or show how to use so it will be easier for people who want to store data..

@arkeros

Jaikant:
@arkeros  Not sure of the difference between the User and the UserLogin table. The both seem to shows the time when the login was created and last time it was updated (last login time?).

PR #546, opened 2016-04-06
cleaner

zwhitchcox:
This has the same effect

---
This change is [Reviewable](https://reviewable.io/reviews/kriasoft/react-starter-kit/546)


koistya:
Thanks for the PR! But.. `.concat(..)` doesn't modify the original array.

PR #590, opened 2016-04-18
Fix: AccessToken from facebook saved as String

sergiufalcusan:
Postgress throws an error because facebook is not returning an Integer.


koistya:
@sergiufalcusan thanks!

PR #601, opened 2016-04-20
update several deps for Redux branch

tomByrer:
Updated deps that didn't fail `npm start/test/run build/run lint`.  A few I had to revert back due to this error:
``` bash
npm WARN eslint-config-airbnb@7.0.0 requires a peer of eslint-plugin-jsx-a11y@^0.6.2 but none was installed.
npm WARN eslint-config-airbnb@7.0.0 requires a peer of eslint-plugin-react@^4.3.0 but none was installed.
```
browser-sync v2.12.3 is unreleased yet
& you know about JSCS ;)


langpavel:
Hey, latest released browser-sync is really 2.12.3. @koistya  why you don't use semver `^` ranges in package.json? 
```
$ npm show browser-sync 
{ name: 'browser-sync',
  description: 'Live CSS Reload & Browser Syncing',
  'dist-tags': { latest: '2.12.3' },
  ...
```

koistya:
@langpavel sometimes a tiny version update may break your project, npm packages tend to be very fragile in nature, so it's safer to have strict versions on "dependencies" section of `package.json` I believe.

tomByrer:
oops didn't see v2.12.3 at [Github](https://github.com/BrowserSync/browser-sync/releases); I just looked at the big bold release.  Ty for `npm show` tip.

I tend to use _._.x versions; patches almost never break, minor versions sometimes do.  Not everyone has the same opinion on SemVer.

tomByrer:
(BTW, I kept most `^` versions except one.

langpavel:
I understand you @koistya. My point is to minimize need for PR of this kind. If something breaks, Travis and greenkeeper should catch it.

langpavel:
For production there should be [`shrinkwrap.json`](https://docs.npmjs.com/cli/shrinkwrap)

langpavel:
@tomByrer @koistya I believe this is not actual any more, is it?

PR #603, opened 2016-04-21
Redux store injection refactored

langpavel:
`App` is now more close to desired future `Layout` component, at least as on `master` branch (code in `src/components/App/` has identical behavior with `master`)


langpavel:
I'm closing this because it's not really useful. I better another approach

PR #613, opened 2016-04-22
Remove routing related boilerplate; rename App to Layout

koistya:
<p align="center"><b>[ early preview, please share your feedback ]</b></p>

- [x] Remove routing related boilerplate code from `server.js` / `client.js`
- [x] Rename `App` component to `Layout`, remove context related stuff from it
- [x] Allow to return metadata from route actions alongside the React component to render
- [x] Being able to pass context vars to the router / React renderer, that becomes available from inside route handlers (actions) and all the React components

#### `server.js` - server-side rendering

```js
import express from 'express'
import createRouter from 'universal-router/react';
import routes from './routes';

const app = express();

app.use(createRouter(routes, {
  context: req => ({
    store: createStore(...) // Create Flux/Redux/Relay store
  })
});

app.listen(process.env.PORT || 3000);
```

#### `client.js` - client-side bootstrap code / entry point

```js
import createRouter from 'universal-router/react';
import routes from './routes';

createRouter(routes, {
  context: {
    store: createStore(...) // Create Flux/Redux/Relay store
  }
});
```

#### `routes/index.js` - app routes (plain JavaScript)

```js
export default {
  path: '/',
  action() { .. }
  children: [
    {
      path: '/posts'
      action() { ... }
    },
    {
      path: '/posts/:id'
      action() { ... }
    },
    {
      path: '/admin',
      action() { ... },
      children: [
        { path: '/users', action() { ... } },
        { path: '/posts', action() { ... } },
      ]
    }
  ]
}
```

#### `routes/login/index.js` - a route example

```js
import LoginForm './LoginForm';
import Layout from '../../components/Layout';

export default {
  path: '/login',
  action() {
    return {
      title: 'Log In',
      component: <Layout showHeader={false}><LoginForm /></Layout>
    };
  }
}
```

For review @frenzzy, @langpavel, @Bogdaan, @/all

P.S.: I havn't pushed the updated version of `universal-router` just yet. It should be ready soon.


src/server.js
```diff
@@ -79,41 +78,15 @@ app.use('/graphql', expressGraphQL(req => ({
...
-    next(err);
-  }
-});
+app.use(createRouter(routes, {
+  template: Html,
+  context: (req) => ({
+    userId: req.user && req.user.id || null,
+    store: {}, // TODO: Initialize Flux/Redux store
+  }),
+}));
```
langpavel: This will pass all HTTP methods, not only GET which is fine, but routes should be aware of method and potentially payload... I don't know implementation of createRouter.

src/server.js
```diff
@@ -79,41 +78,15 @@ app.use('/graphql', expressGraphQL(req => ({
...
-
-    res.status(statusCode);
-    res.send(template(data));
-  } catch (err) {
-    next(err);
-  }
-});
+app.use(createRouter(routes, {
+  template: Html,
+  context: (req) => ({
```
langpavel: Can this be async?

langpavel:
LGTM but I wish to play with this to discover it more deeply. Please, ping me when `universal-router` will be published.

src/routes/error/index.js
```diff
@@ -8,20 +8,23 @@
...
-    );
+  action({ error }) {
+    return {
+      status: error.status || 500,
+      title: error.status === 404 ? 'Page Not Found' : 'Error',
+      component: (
+        <Layout error={error}>
+          <ErrorPage error={error} />
+        </Layout>
+      ),
```
frenzzy: It is not clear why you need Layout here, inside which you just render children depending on the error prop.

As I understand it is made for transfer insertCss contextTypes to children, but maybe it is better to add wrapper on router level which can add different contextTypes like insertCss, reduxStoreProvider, dndContext etc?

koistya: @frenzzy yep, right. The Layout component might not be needed for the Error page, need to remove it.

src/components/Html.js
```diff
@@ -0,0 +1,42 @@
...
+      <meta name="description" content="" />
+      <meta name="viewport" content="width=device-width, initial-scale=1" />
+      <link rel="apple-touch-icon" href="apple-touch-icon.png" />
+      <style id="css" dangerouslySetInnerHTML={{ __html: css }}></style>
+    </head>
+    <body>
+      <div id="container" dangerouslySetInnerHTML={{ __html: content }} />
+      <script src={assets.main.js} />
+      {
+        process.env.NODE_ENV === 'production' &&
```
frenzzy: I think it is better to depend here on `analytics.google.trackingId` because you can simply omit the `GOOGLE_TRACKING_ID` environment variable where you do not need statistics.

frenzzy:
PR looks very promising!
Excited to see `universal-router` updates :)

langpavel:
@koistya please link PR for universal-router here so it's connected-related changes. Thanks!

package.json
```diff
@@ -21,7 +21,7 @@
     "front-matter": "2.0.7",
     "graphiql": "0.7.0",
     "graphql": "0.5.0",
-    "history": "2.1.0",
+    "history": "^3.0.0-2",
```
frenzzy: do not forget to remove `^`

langpavel: Actually.. Is locking down dependencies good for us? I know, semver don't work on some packages and history 3.0.0-2 is just prerelease..

langpavel:
@koistya @frenzzy what happens to this?

koistya:
@langpavel it might be a better idea to split that into multiple PRs, for example in the first PR - extract Layout component from the components/App and use that to style website pages/screens. E.g. 

```js
class AboutPage extends React.Component {
  render() {
    return (
      <Layout title="About Us" breadcrumbs="Home > About">
        <p>Hello, world!</p>
      </Layout>
    );
  }
}
```

langpavel:
I have Provider component in react-intl branch. This is common for whole app. Purpose is to provide all context for all HoCs use in project.

Next separation should be Layout. But I'm confused here. Should be Layout be special? Everything page is... It is component. Layout is component in routes. It is necessary to have special Layout component? I'm not sure because this role is fulfilled in router

langpavel:
In react-router you can subdivide granularity of url resolution in components. This can go complicated but it is powerful..

fchienvuhoang:
I remember, I don't know how to hidden header or footer from app component to make a login or something else, although tried some ways. And I have to create new a layout for do it :)

langpavel:
@koistya I think this is outdated. Closing

PR #622, opened 2016-04-25
Fix linting errors

langpavel:
Sorry about that


langpavel:
@koistya You should not merge if Travis not pass... Don't trust me too much :smile:

PR #626, opened 2016-04-26
Documentation: How to Configure Facebook Login

tobiasmuehl:

PR #627, opened 2016-04-26
Documentation: How to Configure Facebook Login

tobiasmuehl:
Better version of PR #626

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "normalize.css": "4.1.1",
     "passport": "0.3.2",
     "passport-facebook": "2.1.0",
+    "postgraphql": "^1.2.0",
     "pretty-error": "2.0.0",
     "react": "15.0.1",
     "react-dom": "15.0.1",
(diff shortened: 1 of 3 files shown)


koistya:
@tobiasmuehl thanks for the PR! I see there is a new npm dependency in the commit, is it on purpose?

package.json
```diff
@@ -30,6 +30,7 @@
     "normalize.css": "4.1.1",
     "passport": "0.3.2",
     "passport-facebook": "2.1.0",
+    "postgraphql": "^1.2.0",
```
langpavel: Great tip, I really fall in love with [`postgraphql`](https://github.com/calebmer/postgraphql#readme) package, but it does'n belongs here.

tobiasmuehl: The new NPM dependency is not on purpose and only found it's way in the PR
by accident

Pavel Lang <notifications@github.com> schrieb am Mi., 18. Mai 2016, 02:50:

langpavel: Please @tobiasmuehl, remove this.. Thanks!

PR #650, opened 2016-05-13
[feature/redux] Add support for helpers in redux-thunk actions

langpavel:
***Work in Progress***

Support for isomorphic `fetch` and `graphqlRequest`
helpers in redux-thunk action creators.

Extracted (and refactored) from my private work for @bravo-kernel. Not tested in this form.

Proposed usage:
```js
// src/actions/auth/logout.js
import {
  LOGOUT,
} from '../../constants';

const query = `
query {
  logout(logout: true)
}`;

export function logout() {
  return async (dispatch, _, { graphqlRequest }) => {
    await graphqlRequest(query);
    dispatch({ type: LOGOUT });
    return true;
  };
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "react-dom": "15.0.1",
     "react-redux": "^4.4.5",
     "redux": "^3.4.0",
-    "redux-thunk": "^2.0.1",
+    "redux-thunk": "^2.1.0",
     "sequelize": "^3.21.0",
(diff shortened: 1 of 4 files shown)


langpavel:
Thanks! :-)

bjackson:
How do I use these helpers?

langpavel:
@bjackson See `graphqlRequest` helper on top of this PR

PR #653, opened 2016-05-14
[feature/react-intl] Extract `description` into messages files

langpavel:
Extract `description` from `defineMessages` and write these into `src/messages/`...

This is important for translators.

@koistya Ready to merge, don't squash please ;-)


langpavel:
@koistya I believe tests will pass now when you re-run on Travis ;-)

langpavel:
Thanks!

PR #656, opened 2016-05-16
Change test scripts

PepijnSenders:
- Add quotes to mocha glob
- Change watcher to extend from the regular test call

There's a problem where NPM doesn't read the glob argument properly, read some about that [here](https://github.com/npm/npm/issues/10481).

Own example:

```bash
$ npm run test

> ./node_modules/mocha/bin/mocha ./test/**/*Spec.js --require test/setup.js --compilers js:babel-register

  environmentSpec
    isProduction
      ✓ should return true if production
      ✓ should return false if not production


  2 passing (32ms)
```

```bash
$ npm run test                                                                

> ./node_modules/mocha/bin/mocha './test/**/*Spec.js' --require ./test/setup.js --compilers js:babel-register

  ServerSpec
    ✓ should run
    ✓ should set server values (58ms)
    ✓ should use pretty error

  environmentSpec
    isProduction
      ✓ should return true if production
      ✓ should return false if not production


  5 passing (100ms)
```


koistya:
@PepijnSenders thanks!

PR #658, opened 2016-05-17
fix: let mocha run every test suit in src

na57:


na57:
sorry, a mistake happened.

PR #662, opened 2016-05-18
Add support for plain CSS (.css) files

koistya:
- [x] Add a loder for `.css` files in `tools/webpack.config.js`
- [x] Convert SCSS to plain CSS, e.g. `Button.scss => Button.css`
- [x] Replace `precss` with `sass-loader` for `.scss` files

##### Example

```js
import React from 'react';
import b from 'bootstrap/bootstrap.scss';
import s from './Button.css';

function Button() {
  return (
    <button className={s.btn}>Click me</button>
  ); 
}

export default Button;
```

##### Questions

1. Should RSK provide configs for both CSS and Sass/SCSS files or just for CSS?
2. Should RSK include example styles in CSS or SCSS format?


langpavel:
> Should RSK provide configs for both CSS and Sass/SCSS files or just for CSS?

Yes.

> Should RSK include example styles in CSS or SCSS format?

Why not both? But plain CSS wins

langpavel:
Do you fix stylelint issues?

PR #663, opened 2016-05-18
[feature/redux] Fix: Redux helpers client configuration

langpavel:
Not sure if this can be source of @mika75 trouble..


mika75:
@langpavel, this solve the problem thanks

PR #678, opened 2016-05-30
fix(Isomorphic Style Loader): Add support to load multiple styles

dhyegofernando:
More info at [kriasoft/isomorphic-style-loader#37](https://github.com/kriasoft/isomorphic-style-loader/issues/37)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -16,7 +16,9 @@ import history from './core/history';
 import { addEventListener, removeEventListener } from './core/DOMUtils';
 
 const context = {
-  insertCss: styles => styles._insertCss(), // eslint-disable-line no-underscore-dangle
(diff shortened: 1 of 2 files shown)


Everettss:
This solution is not fully working.
There is problem with removing styles when they are no longer referenced by UI components.
More details here: [isomorphic-style-loader/issues/45](https://github.com/kriasoft/isomorphic-style-loader/issues/45)
Possible solution:

```javascript
insertCss: (...styles) => {
    const removeCss = styles.map(style => style._insertCss());
    return () => {
        removeCss.forEach(f => f());
    };
},
```

dhyegofernando:
@Everettss thank you for figured it out. I've just fixed.

PR #683, opened 2016-06-01
Fix OccurrenceOrderPlugin spelling

athomann:


koistya:
I assume this will work only after upgrading to Webpack 2.0

koistya:
@athomann thanks for the PR! It seems like there is a conflict, would you be able to resolve it?

athomann:
Fixed

PR #688, opened 2016-06-05
Add react-helmet to substitute context.setTitle/setMeta

awesomejerry:
By using react-helmet instead of self-implemented setTitle and setMeta, we're able to put more control over head tag.

https://github.com/nfl/react-helmet

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "pretty-error": "2.0.0",
     "react": "15.1.0",
     "react-dom": "15.1.0",
+    "react-helmet": "^3.1.0",
     "sequelize": "3.23.2",
     "source-map-support": "0.4.0",
     "sqlite3": "3.1.4",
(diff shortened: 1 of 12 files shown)


joonhocho:
+1

langpavel:
Helmet is good solution at first look. But when you need to update deeply nested hierarchy of components it can introduce some not cleanly deterministic behaviors. Title and metas should be determined at most higher levels of component hiearchy as possible..

rdewolff:
So we'll keep the current behaviour for now @langpavel ?

langpavel:
Helmet have side effect when (re)rendering component. Helmet allows cascading but this should behave strange when you update component in the middle of hierarchy. I don't know if this is fixed... (Or if this was caused by bad `shouldComponentUpdate` implementation in my code)

So I'm for keeping imperative style for title except someone will proof that using Helmet is really without this unwanted side effects..

ulani:
@awesomejerry  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #691, opened 2016-06-09
Change db string length

zzh8829:
Using mysql with the current configuration will cause
```ERROR 1071 (42000): Specified key was too long; max key length is 767 bytes```

See http://stackoverflow.com/a/16820166 for the change
I think we should make this project compatible with other databases such as mysql.

### DIFF
--- a/src/data/models/User.js
+++ b/src/data/models/User.js
@@ -19,7 +19,7 @@ const User = Model.define('User', {
   },
 
   email: {
-    type: DataType.STRING(256),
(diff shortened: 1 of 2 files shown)


koistya:
@zzh8829 looks good, thanks!

arvigeus:
Do we need explicitly STRING(255)? Default size for STRING in sequelize is 255 chars. Length can be omitted in this case, in my opinion.

PR #698, opened 2016-06-13
Feature/redux

EdisonDignos:


langpavel:
Thank you @EdisonDignos for initiating this. But this will not going to happen. Not now. Some of us are waiting for Relay and another competing technologies.
As new direct collaborator to this awesome project I have much more freedom now and I promise I will keep feature branches in sync with master so if you want, follow this branch. It will be maintained.
Thanks

PR #703, opened 2016-06-15
Add eslint global-require exception

DominikeCruz:
Add eslint global-require exception, based in 'src/server.js' file.

The 'feature/react-intl' branch also has this problem.


DominikeCruz:
I forgot to edit the 'src/serverIntlPolyfill.js' file.

PR #707, opened 2016-06-16
Update stylelint-config-standard

DominikeCruz:
When I run 'npm run lint', this warning is shown:

Deprecation Warning: 'number-zero-length-no-unit' has been deprecated, and will be removed in '7.0'. Use 'length-zero-no-unit' instead. See: http://stylelint.io/user-guide/rules/length-zero-no-unit/

update package to remove it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "redbox-react": "^1.2.6",
     "sinon": "^2.0.0-pre",
     "stylelint": "^6.5.1",
-    "stylelint-config-standard": "^8.0.0",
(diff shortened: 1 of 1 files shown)


frenzzy:
but need to update `stylelint` version too
> [6.6.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#660)
> Deprecated: number-zero-length-no-unit. Use length-zero-no-unit instead.

DominikeCruz:
Thanks for the feedback. I edited it.

PR #711, opened 2016-06-17
Remove jade dependency

frenzzy:
This is a breaking change, please create a tag before merge


immuzov:
Thank you for this PR. It's very helpful for me.

src/components/Html.js
```diff
@@ -0,0 +1,42 @@
...
+      </body>
+    </html>
+  );
+}
+
+Html.propTypes = {
+  title: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  style: PropTypes.string.isRequired,
+  script: PropTypes.string,
```
immuzov: Why script didn't required?

frenzzy: I disabled the script in case of server side errors, otherwise the client.js bundle redraws the page without any errors and the developer can fail to notice it.

To remedy this, the server should return the current state to the client, but this is another task)

immuzov: @frenzzy I'm understood, thank you!

PR #728, opened 2016-07-06
Fix for failing App.js test

jtiscione:
the App.test.js in the feature/redux branch was complaining that nobody's giving the App component a redux store property; I added a couple lines to App.test.js and App.js.


langpavel:
This can still be an issue on current feature branches but your PR is outdated and tests are passing with good warning. :-)

Anyway Thanks!

PR #740, opened 2016-07-13
Make the 'isomorphic' term explicit. Suggest React Static Boilerplate.

jdelafon:
For people unaware of the technology, "react-starter-kit" sounds like the thing to download to get a simple React app. You get all the contrary, because you did not go read what the "isomorphic" means in the description of the kit. It looks like React Static Boilerplate is the better choice in this case.
This PR defines the term "isomorphic" in one line and suggests the alternative in case the user is mistaken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,11 +3,16 @@
 > [React Starter Kit](https://www.reactstarterkit.com) is an opinionated
 > boilerplate for web development built on top of [Node.js](https://nodejs.org/),
 > [Express](http://expressjs.com/), [GraphQL](http://graphql.org/) and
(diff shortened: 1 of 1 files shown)


langpavel:
LGTM

koistya:
@jdelafon thanks for the PR, I have added a couple of links explaining isomorphic and single-page apps into the README file, plus a comparison table that explains the difference between multiple boilerplates. Do you think that serves the purpose? Or, you still prefer including the definition of "isomorphic" apps into README.md?

jdelafon:
Nice, the table very complete and one cannot miss it.

PR #747, opened 2016-07-15
Update tools/run.js to check if files are run directly from Node.js

emjaksa:
Just a small enhancement to check if the file was run directly from Node.js.


koistya:
@emjaksa looks good, thanks! I'm wondering, why there is a conflict..

emjaksa:
Thanks, should be resolved now.

PR #750, opened 2016-07-15
fix link

guillaumevincent:
there is an error in eact-static-boilerplate link


koistya:
@guillaumevincent thanks!

PR #769, opened 2016-07-26
Comments in router, to avoid common WTF moment

langpavel:


langpavel:
@koistya I'm not sure if this is best way, condition may be extended, but it is really bad to break routing if returned status is different form 200...

koistya:
Good comments!

PR #697, opened 2016-06-12
Update build scripts

frenzzy:
**Do not merge!**
Work in progress..
- Do not bundle server-side code by [webpack](https://github.com/webpack/webpack), only [transpile](https://en.wikipedia.org/wiki/Source-to-source_compiler) via [babel](https://babeljs.io/) (this is the main goal ⚽ )
- Parallel build tasks execution, improved build performance
- Replace [React Transform](https://github.com/gaearon/babel-plugin-react-transform) with [React Hot Loader](https://github.com/gaearon/react-hot-loader) v3, closes #588
- New task: deploy static version to [GitHub Pages](https://pages.github.com/)
- Use original [webpack-dev-middleware](https://github.com/webpack/webpack-dev-middleware) instead of patched [webpack-middleware](https://github.com/kriasoft/webpack-middleware)
- Reuse the result of file and style loaders in the server-side code

TODO:
- Add watch mode to the `copy` and `compileServer` tasks

PR #774, opened 2016-07-29
Fix webpack configuration, server side code splitting

langpavel:
This fixes invalid behavior of `require.ensure` in server code

### DIFF
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -261,7 +261,9 @@ const serverConfig = extend(true, {}, config, {
   entry: './server.js',
 
   output: {
-    filename: '../../server.js',
+    path: path.resolve(__dirname, '../build'),
+    filename: 'server.js',
+    chunkFilename: 'server.[name].js',
     libraryTarget: 'commonjs2',
   },
 


langpavel:
@koistya Please merge, Thanks!

koistya:
Maybe a better approach would be to add `webpack.optimize.LimitChunkCountPlugin({maxChunks: 1})` for the server-side bundle?

langpavel:
> Maybe a better approach would be to add webpack.optimize.LimitChunkCountPlugin({maxChunks: 1}) for the server-side bundle?

I'm not sure if this is really needed because server is long living process and module is loaded maximally once during lifecycle..

My use case is split out administration part of website because it is not needed for all users.
For some users I wish forbid loading this scripts if user have no permission in future on the client.

Also in a cluster deployment it is possible that some chunks will not be loaded ever...

tools/webpack.config.js
```diff
@@ -261,7 +261,9 @@ const serverConfig = extend(true, {}, config, {
   entry: './server.js',
 
   output: {
-    filename: '../../server.js',
+    path: path.resolve(__dirname, '../build'),
+    filename: 'server.js',
```
frenzzy: Now links to external assets like images fonts and sounds will not work and/or will be stored twice in `build/assets/` and `build/public/assets/` folders

langpavel: Can we add some more configuration to fix this?

frenzzy: I think that the server side part just should not be bundled by webpack #697

langpavel: Ahhh, but it is. And it solved lot of problems with isomorphic stuff. But I need code splitting..

frenzzy: With this approach we will still use webpack + code splitting for client-side only. For server side we will use [the beautiful node.js modular approach](https://nodejs.org/api/modules.html) + npm

langpavel: @frenzzy I don't want branch code for client and server. I'm try many blind paths to support `require.ensure` in isomorphic code but this is best I found (and only working BTW)

frenzzy: #697 does not break isomorphic code, it only excludes Webpack from the build process of server code. Webpack's `require.ensure` not needed on the server because the server has access to all application files immediately.

..to strip `require.ensure` from the server code we can use [babel-plugin-remove-webpack](https://github.com/knpwrs/babel-plugin-remove-webpack)

langpavel: @frenzzy Ok, ok, but #697 is not ready yet..

PR #776, opened 2016-07-29
[feature/redux] Isomorphic redirect

langpavel:


GalGavu:
@langpavel 
1. It works properly in component.
2. The same goes for route.
3. About 404 status code. You wrote: "just push state to history with http status code" in gitter -> I cannot find any documentation about pushing status code in history library. Do you have an example how to do that?

langpavel:
Hi @GalGavu. It's undocumented and not implemented.. It's only advice ;-)

PR #781, opened 2016-07-30
Dont include same style multiple times . Fix #780

### LINKED ISSUE #780: When Same component is repeated , Same Styles getting inserted multiple times in the page.
While Rendering the component on server-side, When Same component is repeated , Same Styles getting inserted multiple times in the page. This is because we are inseting styles in the array without checking if the same style already exists. 

Thus document size becoming huge for a page.

sap9433:


sap9433:
Can some one please merge this . We have used this code in production , and it works fine .

koistya:
@sap9433 thanks, much appreciated!

PR #793, opened 2016-08-04
Example: google-map-react

langpavel:
DO NOT MERGE. This is only an example

See steps commit by commit how to create new page with google-map-react


langpavel:
Outdated.

PR #799, opened 2016-08-11
Checks if bootstrap fonts folder exists

clayrisser:
Some versions of bootstrap do not come with a fonts directory. This fix allows such versions of bootstrap to be used with this branch.

### DIFF
--- a/tools/copy.js
+++ b/tools/copy.js
@@ -12,6 +12,7 @@ import gaze from 'gaze';
 import Promise from 'bluebird';
 import fs from './lib/fs';
 import pkg from '../package.json';
+import _fs from 'fs';
 /**
  * Copies static files such as robots.txt, favicon.ico to the
  * output (build) folder.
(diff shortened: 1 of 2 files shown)


langpavel:
Can you please follow code style, run `npm run lint`, then I can merge this

tools/copy.js
```diff
@@ -12,19 +12,26 @@ import gaze from 'gaze';
 import Promise from 'bluebird';
 import fs from './lib/fs';
 import pkg from '../package.json';
+import _fs from 'fs';
```
langpavel: Alter lib/fs, add stat function here, do not use underscore prefix

tools/copy.js
```diff
@@ -12,19 +12,26 @@ import gaze from 'gaze';
...
-    ncp('node_modules/bootstrap/dist/css', 'build/public/css'),
-    ncp('node_modules/bootstrap/dist/fonts', 'build/public/fonts'),
+  var ncpArray = [
     ncp('src/public', 'build/public'),
     ncp('src/content', 'build/content'),
-  ]);
+    ncp('node_modules/bootstrap/dist/css', 'build/public/css')
+  ];
+  try {
+    _fs.statSync('node_modules/bootstrap/dist/fonts');
```
langpavel: Use await here and async version of stat

langpavel:
@jamrizzi Any progress?

jamrizzi:
Hey, so the reason I used _fs was because fs was previously defined, and it was using an implementation that doesn't seem to work properly for file streams. I've actually done some customization on this starter kit, and reimplemented bootstrap in a different way, a way that would make this hacky fix pointless. I basically converted all CSS to SCSS and loaded the bootstrap 4 SCSS files in so it can be customized and compiled by webpack. I may also implement redux with it as well. This starter kit is great, but it is missing 3 fundamental pieces, sass, bootstrap, and redux.

Really what might be better is to create a branch of this project when I am finished with it. I'll also create some simple instructions that explain how to replace bootstrap 4 with bootstrap 3, in case someone doesn't want to be using a beta project.

langpavel:
@jamrizzi Great to hear! I give a try to bootstrap4 but I give it up, because I didn't have more time.

If you share some pieces of code, update wiki or write blog post, we will honor your work in README :-)

ulani:
@clayrisser  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #642, opened 2016-05-07
Add --static flag to build script which allows to generate static html for specific routes

frenzzy:
```shell
$ npm run build -- --static --release
```
Build folder with static html files

refs #61 #156 #260 #634 #635

Issue #634, opened 2016-04-30
Q: how to switch off server-side rendering?


Dmitry-N-Medvedev:
I desperately need this functionality to be able to produce static site.

PS: I am aware of https://github.com/koistya/react-static-boilerplate

koistya:
I think @frenzzy is working on adding `--static` flag to the build script, making it generate only the static portion of the site. But that's not done yet.

langpavel:
Why somebody wants to switch off server side rendering???

Dmitry-N-Medvedev:
@langpavel,
1. react-starter-kit is just awesome for development. awesome.
2. big apps is not always an option. oftentimes, one has to implement just a simple app with few screens that must run on the client-side.

PR #833, opened 2016-09-05
Enables to toggle server side rendering when running the server. 

hackhat:
Configuration has a new key named `serverSideRendering` that can be toggled between `true` and `false`. Related to: #642, #634

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -11,6 +11,7 @@
 
 export const port = process.env.PORT || 3000;
 export const host = process.env.WEBSITE_HOSTNAME || `localhost:${port}`;
+export const serverSideRendering = true;
 
 export const databaseUrl = process.env.DATABASE_URL || 'sqlite:database.sqlite';
 
--- a/src/server.js
+++ b/src/server.js
@@ -27,7 +27,7 @@ import models from './data/models';
 import schema from './data/schema';
 import routes from './routes';
 import assets from './assets'; // eslint-disable-line import/no-unresolved
-import { port, auth } from './config';
+import { port, auth, serverSideRendering } from './config';
 
 const app = express();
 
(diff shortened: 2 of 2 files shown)


frenzzy:
It may be better to just not use `server.js` if you want to disable the SSR?

hackhat:
The issue with that is that you need extra work to setup, and managing an extra index.html page. Probably could be rendered once and then served. Then you would need another server config to serve this index.html page including all the static files + hot reload... 
I think this would help to get started for now, what do you think?

frenzzy:
SSR in RSK already configured and allows to do following:
- The code written in one place will work on the client and on the server.
- DEV: the client-side and server-side hot reload just works.
- PROD: website content is visible to search engines like Google, the page loads faster.
- PROD: slow react SSR can be resolved by [caching](https://en.wikipedia.org/wiki/Web_cache) or by compiling [server responses to static files](https://github.com/kriasoft/react-starter-kit/pull/642) during build process.

So, I do not understand your benefits.
If you still consider that you do not need SSR just look towards [RSB](https://github.com/kriasoft/react-starter-kit#comparison).

sauravskumar:
@frenzzy   i think this can be great pull request... since server side rendering as we all know produces some delay... (since ajax has to be performed and wait for rendering to happen after that)... if this pull request can be modified such to perform client side rendering in case of real user and ssr for google crawl bot then it would provide the user with a better user experience in terms of lower page load times (since the app does not need to perform ajax and rendering on server side)... and it would also be good for seo since ssr can be performed in case of google bot... I see many major companies taking this approach now a days such as [flipkart](https://www.flipkart.com/)... 

Great pull request

+1

hackhat:
@frenzzy This is entirely for development reasons; Sometimes you get server side errors and the whole app doesn't start or behaves unexpectedly; If is just client side rendering, you can debug errors easily; At least that is what I've noticed.

hackhat:
Another reason, that happened just now, is that sometimes there is an error and the server doesn't even start, in this case I will disable server side rendering and investigate the issue on the client side.

src/config.js
```diff
@@ -11,6 +11,7 @@
 
 export const port = process.env.PORT || 3000;
 export const host = process.env.WEBSITE_HOSTNAME || `localhost:${port}`;
+export const serverSideRendering = true;
```
langpavel: Use `export const disableServerSideRendering = !process.env.DISABLE_SERVER_SIDE_RENDERING;`

src/server.js
```diff
@@ -27,7 +27,7 @@ import models from './data/models';
 import schema from './data/schema';
 import routes from './routes';
 import assets from './assets'; // eslint-disable-line import/no-unresolved
-import { port, auth } from './config';
+import { port, auth, serverSideRendering } from './config';
```
langpavel: `disableServerSideRendering`

src/server.js
```diff
@@ -84,31 +84,30 @@ app.use('/graphql', expressGraphQL(req => ({
 // -----------------------------------------------------------------------------
 app.get('*', async (req, res, next) => {
```
langpavel: Disable SSR on router level, not inside handler, add default app loader

hackhat:
Thanks @langpavel but I'm not sure if is worth investing more time fixing the conflicts and adding your suggestions if it won't be accepted anyway.

langpavel:
Anyway thanks

PR #839, opened 2016-09-06
Add setDescription function in context

fchienvuhoang:
add setDescription to client.js and server.js to active method setDescription in each router:
`state.context.setDescription('some description');`


fchienvuhoang:
@koistya please check this pull :)

frenzzy:
Why do not you want to use a `context.setMeta('description', 'some description')`?

azz0r:
Agree with the above.

fchienvuhoang:
@frenzzy have you tried with that code ? ins't work
err

fchienvuhoang:
it worked, thanks :)

fchienvuhoang:
because i had not understood below code:
`setMeta: (key, value) => (data[key] = value),`
it ok now :D, 
thanks all

PR #867, opened 2016-09-21
Update history dependency (v3 -> v4.2)

### LINKED ISSUE #855: History needs updated
The history React module was updated. Its new implementation is very different than the way it is implemented in this Starter kit. Here is a link with details about the change.

https://github.com/mjackson/history/issues/364#event-787879740

frenzzy:
close #855


frenzzy:
Closing in favor of #869.

PR #869, opened 2016-09-22
Introduce the new App and Layout components. Upgrade `history` module.

koistya:
- [x] Split the `App` component into `App` that's responsible for setting application context and `Layout` that contains layout-related markup and CSS
- [x] Upgrade [`history`](https://github.com/mjackson/history) npm module to v4.x ([changelog](https://github.com/mjackson/history/blob/master/CHANGES.md))
- [x] Upgrade other npm modules to the most recent versions (no breaking changes)
- [x] Remove `core/createHistory.js` in favor of initializing a new history object inside `server.js` and `client.js`
- [x] Set `history` context variable available from inside React components, e.g. `this.context.history.push('/home')`
- [x] Refactor `Link` component to use `this.context.history` instead of `./core/history.js`


langpavel:
Good job! Will review at weekend, to busy now..

koistya:
In general, it’s not a good idea to specify view hierarchy inside the route declarations, it’s better to do that inside React. At a first glance, it might look like code duplication (specifying the outer (container) component in each route/view component), but it’s very simple mental model, very easy to understand how it works and it’s super powerful. You can always remove code duplication by creating another container component, for example, you can create ProductLayout component on top of the basic `Layout` component, and include side menu, breadcrumbs etc. then use that inside your Product route/view component - `render() { <ProductLayout>…</ProductLayout>; }` instead of the basic `Layout` component.

PR #893, opened 2016-10-03
Allow load critical but optional polyfills

langpavel:
This patch allows load heavy polyfills (such as intl) asynchronously and optionally before app run through webpack code splitting feature.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -156,16 +156,20 @@ async function onLocationChange(location) {
   }
 }
 
-// Handle client-side navigation by using HTML5 History API
-// For more information visit https://github.com/mjackson/history#readme
-context.history.listen(onLocationChange);
-onLocationChange(currentLocation);
+export default function main() {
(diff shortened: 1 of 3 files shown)


frenzzy:
Why do you need this separate chunk if you anyway do not launch the application until all "heavy" dependencies will be loaded? All the critical part of the application must be loaded synchronously and only additional staff can be loaded asynchronously which is not required for initial start.

If you just want to split vendor code then you can create a separate entry point for vendors and load them synchronous:
```html
<script src="/assets/vendor.js"></script>
<script src="/assets/main.js"></script>
```

langpavel:
@frenzzy look at `feature/react-intl`. There is heavy polyfill for the `intl` which is only loaded for Safari. It can be completely ignored otherwise.

Main chunk with app is always loaded and immediately executed in most browsers, but delayed in execution after Safari is polyfilled...

langpavel:
Included in `feature/react-intl`

PR #899, opened 2016-10-04
Add empty user reducer and clean import react.children

keremciu:
Unexpected key "user" found in preloadedState argument passed to createStore. Expected to find one of the known reducer keys instead: "runtime", "intl". Unexpected keys will be ignored.


langpavel:
I squash merge this. This should be cherry picked into `feature/redux` too..

PR #901, opened 2016-10-04
Add code splitting example

frenzzy:


langpavel:
Is this really good example? I'm successfully using webpack with require.ensure only and it works great. One example is in feature/react-intl in clientLoader.js.
Is this only optimization?

frenzzy:
Features:

1. It allows to load a chunk synchronously with the initial page loading.
2. It uses Promise to avoid callback hell and simplify updates in the future.
  ```js
  await System.import('./Component')
  ```

3. Promises allow to do parallel asynchronously multiple requests at once.
  ```js
  const [Component, data] = await Promise.all([
    System.import('./Component'),
    fetch('/api/data').then(resp => resp.json())
  ])
  ```

langpavel:
I see `require.ensure` in your PR. I did'n know that `System.import` is supported by our version of WebPack. If yes, you should use it ;-)

I see you added `chunk` into header, this only preloads chunk if app knows then it will be needed. Cannot this trigger double request on slow connections?

frenzzy:
> I see `require.ensure` in your PR. I did'n know that `System.import` is supported by our version of WebPack. If yes, you should use it ;-)

We do not have `System.import` yet, it is a webpack v2 feature. I'm trying to say is that when we are ready to upgrade we just need to do a single find and replace call:
```js
new Promise((resolve) => {
  require.ensure([], (require) => resolve(require('./moduleName').default), 'chunkName');
});
// =>
System.import('./moduleName');
```
...waiting for [chunk name solution](https://github.com/webpack/webpack/search?q=chunk+name&type=Issues) before upgrading to webpack v2.

> I see you added `chunk` into header, this only preloads chunk if app knows then it will be needed. Cannot this trigger double request on slow connections?

Webpack is smart enough that not to make multiple requests for the same chunk if it has been requested earlier (even before main.js).

view-source
network

Experiment details: release build, no cache, 4G network throttling, artificial 5s response delay for `admin` chunk.

src/components/Html.js
```diff
@@ -45,6 +46,7 @@ Html.propTypes = {
   description: PropTypes.string.isRequired,
   style: PropTypes.string,
   script: PropTypes.string,
+  chunk: PropTypes.string,
```
frenzzy: Why you want to load more than one chunk for single page?

langpavel: Because you want for example big component which you want load lazily in some cases but you know pages where is good load it always? For example source editor with syntax support..

langpavel:
Ok @frenzzy, you convinced me..
LGTM
@koistya merge?

PR #903, opened 2016-10-06
Feature/redux

aJoohongKim:
merge the latest of master.
In src/server.js, I couldn't remove history entirely due to store dependency. Removed history from context.

Could you check?


langpavel:
I believe all changes are now merged.

> I couldn't remove history entirely due to store dependency.

What you think by this?

PR #686, opened 2016-06-04
Add static content intl(locale) support

awesomejerry:
related to https://github.com/kriasoft/react-starter-kit/issues/684

support:
```
src/content/about.en.jade
src/content/about/en.jade
```

Issue #684, opened 2016-06-02
static content intl support


awesomejerry:
Hi all, really love this kit for handling so much cool stuff together. After merged with feature/react-intl, I was able to define messages for components, but I wonder if this can be extended to static content under src/content folder.
Possible usage might be:

```
src
--- content
------ about.en.jade
------ about.zh_Hant.jade
```

based on what locales you have defined in src/config.js

langpavel:
You can customize yourself handling of static content loading.
Copy and resolve this checklist:

 * [ ] update `src/data/queries/content.js` to accept and handle `locale` argument (See `src/data/queries/intl.js` as reference).
 * [ ] This argument should be optional, defaulting to `req.language` (optional behavior)
 * [ ] in `src/routes/content/index.js` extend GraphQL query with `locale` argument

AwesomeJerry:
Thanks for the checklist, making it so clear about implementing. Will make a PR when I have the time to resolve it :)

langpavel:
Great, don't forget to make PR against `feature/react-intl` branch and keep backward compatibility ;-)

PR #912, opened 2016-10-13
Add static content intl(locale) support

awesomejerry:
related to #684, #686  

support:

```
src/content/about.en.jade
src/content/about/en.jade
```


langpavel:
Hi @awesomejerry
This code is build on `feature/react-intl` which internally uses Redux, but you implement this through `context`. This is not preferred way of doing fetching in Redux branches. If you do this by redux, you will see that you can entirely drop `localeDetect` HoC which is also not good example of HoC..

You can look at `LanguageSwitcher` and try rewrite `Content` code to use Redux.

I know I should rewrite content fetching too and I probably do this soon but for now I will do not merge this.

Anyway Thanks! And ping me in week if nothing happens here :smiley:

awesomejerry:
The main reason I wrote `localeDetect` is that I don't want to mess up with the original code as much as possible (which is not meant for HoC demonstration apparently), trying to make things happen just inside a new component (that's why the usage of internal state). Furthermore, the locale content is only used by `Component`, so I didn't use redux or even create a new reducer. But hey, making things consistent might be great though. I'll make some changes by the end of the week.

FYI, I'm gonna make `Component` a stateful component in order to detect client side language switch. Dispatch action whenever the locale changes and take the content from `state.content.data`. Please let me know if there's any recommended ways of doing so.

awesomejerry:
Hi @langpavel , I've updated the PR to redux style. Please check it. Thanks!

src/actions/content.js
```diff
@@ -0,0 +1,43 @@
...
+    content(path:$path,locale:$locale) {
+      title, content,
+    }
+  }
+`;
+
+export function getLocaleContent(path, locale) {
+  return async (dispatch, getState, { graphqlRequest }) => {
+    dispatch({
+      type: GET_LOCALE_CONTENT,
```
langpavel: * `FETCH_CONTENT_START`
* add more data to payload, path, locale
* you can make locale optional witch default value from `getState().intl.locale`

src/actions/content.js
```diff
@@ -0,0 +1,43 @@
...
+
+const query = `
+  query($path:String!,$locale:String!) {
+    content(path:$path,locale:$locale) {
+      title, content,
+    }
+  }
+`;
+
+export function getLocaleContent(path, locale) {
```
langpavel: use object for passing in arguments
```js
export function getLocaleContent({ path, locale })
```

src/constants/index.js
```diff
@@ -4,3 +4,6 @@ export const SET_RUNTIME_VARIABLE = 'SET_RUNTIME_VARIABLE';
 export const SET_LOCALE_START = 'SET_LOCALE_START';
 export const SET_LOCALE_SUCCESS = 'SET_LOCALE_SUCCESS';
 export const SET_LOCALE_ERROR = 'SET_LOCALE_ERROR';
+export const GET_LOCALE_CONTENT = 'GET_LOCALE_CONTENT';
```
langpavel: `FETCH_CONTENT_START` note it is very good practice to use suffixes consistently, see for example [react-redux-loading-bar](https://mironov.github.io/react-redux-loading-bar/)

src/reducers/content.js
```diff
@@ -0,0 +1,36 @@
...
+  },
+};
+
+export default function runtime(state = defaultState, action) {
+  switch (action.type) {
+    case GET_LOCALE_CONTENT:
+      return {
+        ...state,
+        isFetching: true,
+      };
```
langpavel:
```js
const key = `${action.payload.locale}:${action.payload.path}`;
return {
  ...state,
  [key]: {
    ...(state[key]),
    isFetching: true,
  },
}
```
`key` should be extracted to standalone function based on `action.payload`

src/reducers/content.js
```diff
@@ -0,0 +1,36 @@
...
+    case GET_LOCALE_CONTENT_SUCCESS:
+      return {
+        ...state,
+        isFetching: false,
+        data: action.payload.data,
+      };
```
langpavel:
```js
return {
  ...state,
  [key]: {
    isFetching: false,
    ...action.payload
  },
};
```

src/routes/content/Content.js
```diff
@@ -7,28 +7,54 @@
...
 }
 
-Content.propTypes = {
-  path: PropTypes.string.isRequired,
-  content: PropTypes.string.isRequired,
-  title: PropTypes.string,
+const mapState = (state) => ({
+  locale: state.intl.locale,
+  data: state.content.data,
+});
```
langpavel:
```js
const mapState = (state, props) => {
  const locale = state.intl.locale;
  const content = state.content[`${locale}:${props.path}`];
  return {
    locale: state.intl.locale,
    ...content
  };
};
```

awesomejerry:
Thanks for the code reviews and suggestions! They're quite helpful.

langpavel:
Thanks @awesomejerry 
I refactored some stuff to agressively cache content on client. Have a look!

Issue #825, opened 2016-08-29
Error "Cannot resolve module" when import a external css file contains `url(image.png)`


manhhailua:
I'm using React-Starter-Kit and trying to integrate with AdminLTE. When I imported a css from AdminLTE I got this error.

RSK always searches for the image as a module from the component folder, in this case `/home/nginx/domains/vps.manhhailua.com/public/src/routes/login`.

```
webpack built a436c24672c3c331acc3 in 7118ms
Child
    Time: 7118ms

    ERROR in ./~/css-loader?{"sourceMap":true,"modules":true,"localIdentName":"[name]_[local]_[hash:base64:3]","minimize":false}!./~/postcss-loader?pack=default!./src/routes/login/Login.css
    Module not found: Error: Cannot resolve module 'blue.png' in /home/nginx/domains/vps.manhhailua.com/public/src/routes/login
     @ ./~/css-loader?{"sourceMap":true,"modules":true,"localIdentName":"[name]_[local]_[hash:base64:3]","minimize":false}!./~/postcss-loader?pack=default!./src/routes/login/Login.css 6:1349-1368

    ERROR in ./~/css-loader?{"sourceMap":true,"modules":true,"localIdentName":"[name]_[local]_[hash:base64:3]","minimize":false}!./~/postcss-loader?pack=default!./src/routes/login/Login.css
    Module not found: Error: Cannot resolve module 'blue@2x.png' in /home/nginx/domains/vps.manhhailua.com/public/src/routes/login
     @ ./~/css-loader?{"sourceMap":true,"modules":true,"localIdentName":"[name]_[local]_[hash:base64:3]","minimize":false}!./~/postcss-loader?pack=default!./src/routes/login/Login.css 6:2797-2819
Child
    Time: 6019ms
            Asset    Size  Chunks             Chunk Names
        server.js  197 kB       0  [emitted]  main
    server.js.map  192 kB       0  [emitted]  main

    ERROR in ./~/css-loader?{"sourceMap":true,"modules":true,"localIdentName":"[name]_[local]_[hash:base64:3]","minimize":false}!./~/postcss-loader?pack=default!./src/routes/login/Login.css
    Module not found: Error: Cannot resolve module 'blue@2x.png' in /home/nginx/domains/vps.manhhailua.com/public/src/routes/login
[...]

Issue #824, opened 2016-08-29
Error "Unexpected token ." while import external css to a component


manhhailua:
I'm using React-Starter-Kit and trying to integrate with AdminLTE. When I imported a css from AdminLTE I got this error.
```
import s from './Login.css';
import s_iCheck from 'admin-lte/plugins/iCheck/square/blue.css';
```
If I change the background url from url(blue.png) to url("./blue.png"), the error "Cannot resolve module" will disappear but the error "Unexpected token ." still exist.

```
/home/nginx/domains/vps.manhhailua.com/public/node_modules/admin-lte/plugins/iCheck/square/blue.css:3
.icheckbox_square-blue,
^
SyntaxError: Unexpected token .
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:513:28)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/external "admin-lte/plugins/iCheck/square/blue.css":1:1)
    at __webpack_require__ (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/webpack/bootstrap cf2ab8e448863bf0ab23:19:1)
    at Object.<anonymous> (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/routes/login/Login.js:15:1)
    at __webpack_require__ (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/webpack/bootstrap cf2ab8e448863bf0ab23:19:1)
    at Object.module.exports.Object.defineProperty.value (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/routes/login/index.js:11:1)
    at __webpack_require__ (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/webpack/bootstrap cf2ab8e448863bf0ab23:19:1)
    at Object.module.exports.Object.defineProperty.value (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/routes/index.js:16:1)
    at __webpack_require__ (/home/nginx/domains/vps.manhhailua.com/public/build/webpack:/webpack/bootstrap cf2ab8e448863bf0ab23:19:1)
```

[...]

Issue #510, opened 2016-03-17
Having issues using CSS


Ehesp:
I've read the other issues submitted here, but I can't get CSS to load.

In an component, I do something like:

```
import s from 'rc-dropdown/assets/index.css';
...
export default withStyles(Dropdown, s);
```

However I get the error: `SyntaxError: Unexpected token ILLEGAL`

I've tried modifying the webpack config to both:

```
      {
        test: /\.css$/,
        loaders: [
          'isomorphic-style-loader',
          `css-loader?${DEBUG ? 'sourceMap&' : 'minimize&'}modules&localIdentName=` +
          `${DEBUG ? '[name]_[local]_[hash:base64:3]' : '[hash:base64:4]'}`
        ]
      },
```
and
```
{
        test: /\.scss$|.css$/,
        loaders: [
          'isomorphic-style-loader',
          `css-loader?${JSON.stringify({
            sourceMap: DEBUG,

            // CSS Modules https://github.com/css-modules/css-modules
            modules: true,
            localIdentName: DEBUG ? '[name]_[local]_[hash:base64:3]' : '[hash:base64:4]',

            // CSS Nano http://cssnano.co/options/
            minimize: !DEBUG,
          })}`,
          'postcss-loader?parser=postcss-scss',
          'sass-loader?sourceMap',
        ],
      },
```

Still getting the same error - what am I supposed to do?

Bogdaan:
Check you css file (rc-dropdown/assets/index.css) for HTTP/https imports.

And change to:
```
{
        test: /\.s?css$/,
        loaders: [
          'isomorphic-style-loader',
          `css-loader?${JSON.stringify({
            sourceMap: DEBUG,
            localIdentName: DEBUG ? '[name]_[local]_[hash:base64:3]' : '[hash:base64:4]',
            // CSS Nano http://cssnano.co/options/
            minimize: !DEBUG,
          })}`,
          'postcss-loader?parser=postcss-scss',
          'sass-loader?sourceMap',
        ],
      },
```

link #499
link #423

Ehesp:
Thanks

I changed it to the above, however I still get the same error. The css file starts off like so:

```
@font-face {
  font-family: 'anticon';
  src: url('//at.alicdn.com/t/font_1434092639_4910953.eot');
  /* IE9*/
[...]

Issue #423, opened 2016-01-28
how to  require CSS file ?


zhaolihang:
require('./index.css');
when build ,I got a error .

then another error happend when I add "css-loader" in webpack.config.js  :
     {
        test: /.css$/,
        loader: 'css-loader',
      },

how to do it ?
thanks!

koistya:
##### `tools/webpack.config.js/module/loaders`

```js
{
  test: /\.css$/,
  loaders: [
    'isomorphic-style-loader',
    `css-loader?${DEBUG ? 'sourceMap&' : 'minimize&'}modules&localIdentName=` +
    `${DEBUG ? '[name]_[local]_[hash:base64:3]' : '[hash:base64:4]'}`
  ]
}
```

##### `src/components/MyComponent/MyComponent.css`
```css
.root { color: red; }
```

##### `src/components/MyComponent/MyComponent.js`
```js
import React, { PropTypes } from 'react';
import withStyles from 'isomorphic-style-loader/lib/withStyles';
import s from './MyComponent.css';

function MyComponent({ name }) {
  return <div className={s.root}>Hello, {name}!</div>
}

MyComponent.propTypes = { name: PropTypes.string.isRequired };

export default withStyles(MyComponent, s);
```

zhaolihang:
Can I do like this: ?
`import './MyComponent.css';
function MyComponent({ name }) {
  return <div className="root">Hello, {name}!</div>
}`

b-gran:
@zhaolihang with the default settings, you can't do 
```
import './MyComponent.css';
function MyComponent({ name }) { return <div className="root">Hello, {name}!</div> }
```

The default css loader changes class names so you can have per-component stylesheets.

`webpack.config.js`
```
`css-loader?${DEBUG ? 'sourceMap&' : 'minimize&'}modules&localIdentName=` +
    `${DEBUG ? '[name]_[local]_[hash:base64:3]' : '[hash:base64:4]'}`
```
That allows you to create a stylesheet for each component, and for example have a `.root` class in each file with no conflicts. The css loader will rename the `.root` class in each file, so there essentially won't be any global classes. 

If you want your components to share css classes, you have a few options.

1) In your component .s?css files, import the .s?css files of other components you wish to inherit from. Then import the stylesheet in your React component and reference the classname using js.

`SomeComponent.scss`
```
...
@import '../App/App.scss';
.aStyleInheritedFromApp {
/* This class inherits from a class in App */
}
```

[...]

PR #920, opened 2016-10-16
Add ability to import external css

### LINKED ISSUE #375: Question: bug with style-loader?
When i work with css background-image source map generate "Blob" url, and images not loaded from dev-server:
  
    Request URL:blob:http%3A//localhost%3A3000/2454cd1b-a4bb-46fa-8d4b-446ac9b2c9a6   

In style-loader i found this note: [output.publicPath](https://github.com/webpack/style-loader/issues/55)

It posible load image from files with current webpack config (with publicPath=/)?

### LINKED ISSUE #771: How do you include a CSS file from node_modules without it prefixing?
When I do ...
```
import 'react-select/dist/react-select.css';
```
Then the css is dumped to the page but with all of the extra stuff in the name, expecting that I'll be using the withStyles method, but it's not my component and doesn't use that method. With how all of this is configured, how can I bypass the renames?

Would it be a good idea to modify this starter kit to have a css loader for anything in the node_modules directory where it doesn't do all the class renaming?

frenzzy:
- Allows to import external css from node_modules (ref #510, #824)
- Allows to properly process `url()` statements (ref #825)
- Allows to import css without css-modules prefixing (close #771)

  ```css
  /* Example: MyTextEditorComponent.css */
  :global {
    @import 'draft-js/dist/Draft.css'; /* external style without prefixing */
  }
  .button { color: red; } /* keep the prefixes for your own style */
  ```

Also closes #375 and #423


kevinchau321:
This feature isn't working for me. I'm trying to the css from a node module called rc-slider ('rc-slider/dist/rc-slider.css') similar to #510 and #771. Using the `:global` import syntax in my component stylesheet doesn't seem to be working. But I have gotten the external css to load using `import s from '!!isomorphic-style-loader!css-loader!rc-slider/dist/rc-slider.css';` or by just linking the css in the html <head> like `<link rel="stylesheet" href="https://npmcdn.com/rc-slider@6.1.1/dist/rc-slider.css" />`.

I'm not really happy with either one of these work arounds so I'm wondering what I'm doing wrong with my global import.

frenzzy:
It happens because of issue with https://github.com/jonathantneal/postcss-partial-import/issues/35
refs #1123, #1144, #1145
As a temp solution you can rollback this plugin to `v2.*`

glenne:
What is the status of this?  kevinchau321 commented on Feb 21 that the :global import did not work.  This is also my experience but fortunately his import s from example did work for me.

frenzzy:
@glenne you just need to import external css inside your component:
```js
import React from 'react';
import withStyles from 'isomorphic-style-loader/lib/withStyles';
import external from 'external.css';
import s from './MyComponent.css';

class MyComponent extends React.Component {
  render() {
    return <div className={s.local}><div className="global">Hi</div></div>
  }
}

export default withStyles(external, s)(MyComponent);
```
refs #1277, #1274

PR #921, opened 2016-10-17
Fixed test for required store props in App Layout

granda:
The Layout component failed the test because the inner App component required a `store` in its `PropTypes`.

Issue has been resolved by utilizing [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) to create a mock store which is then passed into the App component within the Layout test.


langpavel:
@granda Will merge this by hand
Thank You! :tada:

PR #923, opened 2016-10-19
Update getting-started.md

cdvel:


docs/getting-started.md
```diff
@@ -108,7 +108,7 @@ $ npm run build
 or, for a production build:
 
 ```shell
-$ npm run build -- --release
+$ npm run build --release
```
frenzzy: This is incorrect, application will be compiled in development mode in this case.

langpavel:
@cdvel You can add a note explainin why extra `--` is needed. It is obvious that it is not obvious for everyone :smile:

PR #926, opened 2016-10-19
prevent static resources from enumerating routes

Strandedpirate:
Static resources like css, images and scripts should be 404'd upon entering the routing middleware because they should have already been handled by the express static middleware defined higher in server.js. By allowing them to fall through and enumerate all the routes until we reach the end for the notFound route we waste valuable bandwidth and CPU. graphql calls are also being made in these scenarios by the content route causing unnecessary churn. In addition this is a denial of service vector.

**Side note:** it would be awesome if the [universal router](https://github.com/kriasoft/universal-router) allowed "named" routes so rather than doing this:
`require('./routes/notFound').default.action()`
I could instead do this:
`UniversalRouter.getNamed(routes, 'notFound')`

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -84,9 +84,13 @@ app.use('/graphql', expressGraphQL(req => ({
 // Register server-side rendering middleware
 // -----------------------------------------------------------------------------
 app.get('*', async (req, res, next) => {
-  try {
+  /**
+   * Renders the final html
+   * @param res - the response object
+   * @param route - the matched route
+   */
+  const renderPage = (res, route) => {
     const css = new Set();
-
     // Global (context) variables that can be easily accessed from any React component
     // https://facebook.github.io/react/docs/context.html
     const context = {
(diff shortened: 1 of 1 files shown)


src/server.js
```diff
@@ -98,6 +102,47 @@ app.get('*', async (req, res, next) => {
...
+   * @param resourcePath - the path of the requested resource. e.g. /images/lucky.png
+   * @returns {boolean} - true if the path is a static resource; false otherwise
+   */
+  const isStaticResource = resourcePath => {
+    const periodIndex = resourcePath.indexOf('.');
+    /*
+      if the path contains a period and has at least two characters after said period then it is
+      a static resource
+     */
+    if (periodIndex > -1 && resourcePath.substring(periodIndex + 1).length > 2) {
```
ClemensSahs: Is this static detection little to less?

API calls (.json) for extern scripts as one example

ClemensSahs:
this is a side effect in my mind, is there any middleware solution for this?

currently we have a small 37 line function
after this patch 74 line

langpavel:
I think this is pointless.

Static files are looked up before running server side renderer by express.

Client-side router should not be triggered for each request, it depends on your code.

@Strandedpirate I'm wrong in something?

ulani:
@Strandedpirate  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #936, opened 2016-10-25
Add Heroku deployment feature

AntonyKwok:
I've added an all-in-one solution to Heroku deployment and I've updated the docs as well.
Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
     "copy": "babel-node tools/run copy",
     "bundle": "babel-node tools/run bundle",
     "build": "babel-node tools/run build",
-    "deploy": "babel-node tools/run deployToAzureWebApps",
+    "deploy": "babel-node tools/run deployToHeroku",
     "render": "babel-node tools/run render",
     "serve": "babel-node tools/run runServer",
     "start": "babel-node tools/run start"
--- a/README.md
+++ b/README.md
@@ -1,224 +1,55 @@
-## React Starter Kit — "[isomorphic](http://nerds.airbnb.com/isomorphic-javascript-future-web-apps/)" web app boilerplate
+## Build Automation Tools
 
-> [React Starter Kit](https://www.reactstarterkit.com) is an opinionated
-> boilerplate for web development built on top of [Node.js](https://nodejs.org/),
-> [Express](http://expressjs.com/), [GraphQL](http://graphql.org/) and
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -180,7 +180,7 @@
     "copy": "babel-node tools/run copy",
     "bundle": "babel-node tools/run bundle",
     "build": "babel-node tools/run build",
-    "deploy": "babel-node tools/run deployToAzureWebApps",
```
langpavel: Why remove this one? It can be splitted to separate targets `deploy:...`

Andriy-Kulak:
@AntonyKwok I tried running your command for heroku deployment command (yarn run deploy) but I get the following error:
- Failed to detect app matching https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/nodejs.tgz buildpack

This means that heroku doesn't recognize this as node type of app even though I can run the app locally fine. Seems very simple so I don't understand what the problem is. Here were my steps:
1. yarn install
2. set up heroku instance. add node buildpack in heroku
3. configure remote object in deploy.js  for heroku deployment as indicated in the comments
4. yarn run deploy (then I get the error)

Is there anything I am missing here? I am used to simply pushing to remote branches (git push prod master, etc.) so I am not as familiar with the approach used in RSK.

AntonyKwok:
Hi Andriy,

The new RSK deploy command incorporates both GitHub pages and Microsoft Azure deployment options into a single file.

To deploy to Heroku, I've now just modified the beginning of `deploy.js` to configure it with the Heroku URLs:

```javascript
// Heroku
const app = 'app-name';

const remote = {
  name: 'heroku',
  url: `https://git.heroku.com/${app}.git`,
  branch: 'master',
  website: `https://${app}.herokuapp.com`,
};
```

Andriy-Kulak:
@AntonyKwok I follow the instructions but it doesn't work.

I tried doing 'yarn run deploy' and 'git push heroku master' (once I added a remote branch).

I don't understand what to fix. package.json files and yarn files are in the root directory. What else can I check?

```
-----> Failed to detect app matching https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/nodejs.tgz buildpack
       More info: https://devcenter.heroku.com/articles/buildpacks#detection-failure
 !     Push failed
```

PR #977, opened 2016-11-18
Fixes #970: Remove the graphiql module since express-graphql loads this via cdn, …

### LINKED ISSUE #970: Shouldn't graphiql (in browser ide for exploring graphql) and sqlite3 be under devDependencies instead of dependencies in package.json?
Shouldn't graphiql (in browser ide for exploring graphql) and sqlite3 be under devDependencies instead of dependencies in package.json?

buildbreakdo:
Remove the graphiql module since express-graphql loads this via cdn, disable graphiql ide in for production servers. See #970 for discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "fastclick": "1.0.6",
     "fbjs": "0.8.5",
     "front-matter": "2.1.0",
-    "graphiql": "0.7.8",
     "graphql": "0.7.2",
     "history": "4.4.0",
     "isomorphic-style-loader": "1.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,7 +18,6 @@
     "fastclick": "1.0.6",
     "fbjs": "0.8.5",
     "front-matter": "2.1.0",
-    "graphiql": "0.7.8",
```
koistya: Shouldn't it be moved to devDependencies section?

buildbreakdo: express-graphql brings it in via CDN and doesn't allow configuration to leverage the graphiql module. Added to discussion on #970 explaining.

Issue #981, opened 2016-11-20
Copy-pasting `git clone` has extra space.


karlpatrickespiritu:
````cli
Private@KarlEspiritu-PC MINGW32 ~/Development/web
$ git clone -o react-starter-kit -b master --single-branch \ https://github.com/kriasoft/react-starter-kit.git react-shop
Cloning into 'react-shop'...
fatal: I don't handle protocol ' https'
````

langpavel:
Closed by ac5edf5806cccb903966b69061f2c99f2aa9a063

PR #982, opened 2016-11-20
Removed spacing on `git clone` command.

karlpatrickespiritu:
So that copy-pasting in terminal won't include spaces which will throw errors.


karlpatrickespiritu:
#981

langpavel:
Extra spaces in terminal should not be a problem, but I agree this should be changed. There should be notice about feature branches which I actively maintain :-)

langpavel:
This does not look as problem. But I will remove extra line

PR #988, opened 2016-11-22
Fixes #987: Expose state and search props of the history module on the Link compo…

### LINKED ISSUE #987: Expose search(query) and state functionality of history module in Link component
At present the Link component does not expose the search or state properties of the history module. 

From the [history module readme](https://www.npmjs.com/package/history):
```
// Push a new entry onto the history stack.  <-------- We only support this scenario
history.push('/home')             
 
// Push a new entry onto the history stack with a query string 
// and some state. Location state does not appear in the URL. 
history.push('/home?the=query', { some: 'state' })
 
// If you prefer, use a single location-like object to specify both 
// the URL and state. This is equivalent to the example above. 
history.push({
  pathname: '/home',
  search: '?the=query',
  state: { some: 'state' }
})

```
Only the first scenario is supported. This means the Link component cannot query the server, goal is expose state and search (query) so functionality like link to article by id (among many other things) could be built.

```js
<Link to='/article' query={{id: '1234'}}> Some Article </Link>
```

This is a precursor to a pull request demonstrating create, read, update, delete. Trying to do small pull requests.

buildbreakdo:
See #987 for extended discussion. Expose state and search (aka query) properties of `history` module in the Link component. Automatically parameterize shallow objects and optionally allow a string to be used as the query prop.

### DIFF
--- a/src/components/Link/Link.js
+++ b/src/components/Link/Link.js
@@ -39,11 +39,25 @@ class Link extends React.Component {
     }
 
     event.preventDefault();
-    history.push(this.props.to);
+    history.push({
+      pathname: this.props.to,
+      state: this.props.state ? this.props.state : undefined,
+      search: this.props.query ? this.params(this.props.query) : undefined
+    });
   };
 
+  params = (arg) => {
+    if (typeof arg === 'string') {
+      return arg.startsWith('?') ? arg : `?${arg}`;
+    } else if (arg === Object(arg) && Object.prototype.toString.call(arg) === '[object Object]' ) {
+      return '?' + Object.keys(this.props.query).map((key) => (
+        `${encodeURIComponent(key)}=${encodeURIComponent(this.props.query[key])}`
+      )).join('&');
+    }
+  }
+
   render() {
-    const { to, children, ...props } = this.props;
+    const { to, children, query, state, ...props } = this.props;
     return <a href={to} {...props} onClick={this.handleClick}>{children}</a>;
   }
 }
--- a/src/components/Link/Link.js
+++ b/src/components/Link/Link.js
@@ -21,6 +21,11 @@ function isModifiedEvent(event) {
 class Link extends React.Component {
   static propTypes = {
     to: PropTypes.string.isRequired,
+    query: PropTypes.oneOfType([
+      React.PropTypes.string,
+      React.PropTypes.object,
+    ]),
+    state: PropTypes.object,
     children: PropTypes.node,
     onClick: PropTypes.func,
   };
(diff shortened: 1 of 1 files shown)


frenzzy:
Why don't you want to just use an object?
```js
<Link
  to={{
    pathname: '/home',
    search: '?the=query',
    state: { some: 'state' }
  }}
>
  The Link
</Link>
```

buildbreakdo:
@frenzzy Interesting point -- although that does look a lot like configuration objects of old world jQuery plugin development (cringe). For all components instead of listing out props, we could use configuration objects, a single prop entry point. Would argue idiomatic react is listing out the props as attributes. I don't think the `history` module was designed with JSX in mind, and believe that when we are in JSX land we should use JSX code style -- even if the underlying API to be interfaced with is written in configuration object style. It also cuts down on code duplication (think about this as a general rule), you would be writing `this.props.to.pathname`, `this.props.to.search`, `this.props.to.state` all over the place if you had to perform any transformations before plugging the values into a configuration object. That's much more verbose than just say, `this.props.pathname`, `this.props.search`, `this.props.state` (again think about this as a general rule for other components interfacing with APIs that use configuration objects).

Either way, as is, `<Link to={{pathname: '/home'}}> Some Link </Link>` wouldn't work because of the `to: PropType.string.isRequired` check, at a minimum that would have to be updated to something like:

```js
...
    to: PropTypes.oneOfType([
      React.PropTypes.string,
      React.PropTypes.object,
    ]), 
...
```

Thoughts?

koistya:
As an alternative solution, you could either create a URL builder component (might be a good idea especially if you're using named routes) or extend the Link component (e.g. ArticleLink).

```jsx
<Link to={url.article(id)}>{title}</Link>
```

buildbreakdo:
@koistya What about state? Not exactly sure how to get state out of the `history` module, API docs make it obvious how to get it in, but not out. Or is `history` management of state bad practice as it should be in something like a central redux store? Don't know the answer to that.

In the interim, how about the middle of the road approach which @frenzzy suggested? Just allow passing string or object props for the `to` attribute? Makes the pull request more incremental and still achieves the objective of having the full history API exposed. 

@frenzzy Curious -- what are your thoughts about configuration objects versus JSX attributes?

frenzzy:
I think it is much simpler and enough for most web-apps to just always use href string:
```js
<Link to="/products?phones#iphone">iPhone</link>
<Link to={`/products/${productName}?tab=specs`}>Tech Specs</link>
```
..than always type a few verbose jsx props.

Also I can't imagine what exactly data is better to store inside browser history rather than inside url query params or flux/redux/relay state.

But for complex web-apps with named routes maybe it is a good idea to do like so:
```js
<Link to="routeName" params={{ some: 'id', or: 'query' }}>The Link</Link>
```

buildbreakdo:
@frenzzy Like that top example. Closing this as it's already possible to pass params, elegant too. Love string templates. Thanks for the discussion.

PR #990, opened 2016-11-22
Move the public folder to the root directory

frenzzy:
Because the `public` folder is not actually the source code


slackday:
I've followed PRs and many great changes as of late! Thanks @frenzzy @koistya and all others involved. 

How about creating a new release tag with all resent changes after it has been tested and merged?

Sorry if I hijacked this thread.

koistya:
@slackday yep, need to create a new tag after these changes are merged. BTW, what do you think about having data API (GraphQL or REST API + OAuth) as a [separate project](https://github.com/kriasoft/graphql-starter-kit) in order to reduce complexity of this starter kit? Also, it might be easier for dev teams to work on front-end and back-end independently from each other. You would be able to deploy them under the same domain if needed by using a reverse proxy.

slackday:
Great!

My personal preference is to keep it as it is. I think separation right now with `src/data` sub-folder is easy to understand. I don't have any experience using RSK in bigger teams but we've worked a few developers at the same time on a project I'm working on right now and it's no problem.

We deploy under the same domain using a reverse proxy right now without any issues.

I'm afraid the complexity of getting started would be higher if we separate it into two different projects. As this is a starter kit I think it should be a one-stop goto project with everything needed. RSK is great for this!

What issue are people having wanting to have it separate? If a bigger team needs it it would be easy enough for them to do themselves is my opinion.

slackday:
Oh and if you need my opinion on a default backend I definitely think GraphQL should be default. REST with node/react is well documented right now.

If we need to lower the bar for people to get started maybe we should work on better documentation/examples for the RSK+GraphQL.

When I started using RSK React was pretty new to me and GraphQL was really new. The issues I was having were purely understanding these technologies themselves and not related to RSK specifically. I think others may experience the same thing. To which I think solid examples are the way forward.

PR #1022, opened 2016-12-10
Update how-to-use-sass.md

dabrowski-adam:
Just a small typo. In the actual project there is `default` not `defaults`.

### DIFF
--- a/docs/recipes/how-to-use-sass.md
+++ b/docs/recipes/how-to-use-sass.md
@@ -52,7 +52,7 @@ const config = {
   ...
   postcss(bundler) {
     return {
-      defaults: [
+      default: [
         ...
       ],
       sass: [


koistya:
Good catch. I think it would make sense to change it to `defaults` in the project. This way when you don't specify the list of plugins explicitly, it will use `defaults` section (e.g. `css-loader?pack=defaults` is the same as `css-loader`).

poorpaddy:
Would be even better to see SASS included alongside the current CSS loader by default and make this document unnecessary.

ulani:
@dabrowski-adam  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1034, opened 2016-12-16
Removal of gaze in extractMessages

piteer1:
@langpavel Gaze was removed from dependencies, so the intl branch is currently broken. Added also `ignoreInitial: true` like in master branch.


langpavel:
@piteer1 Thank you!

PR #1057, opened 2016-12-31
Remove data API endpoint from the boilerplate

koistya:
**Note**: This is work-in-progress. Do not merge!

- [x] Remove `GraphQL` endpoint, `Sequelize`, `sqlite3`, `JWT` etc.
  This should reduce the complexity of the starter kit. Many developers actually prefer having front-end and API endpoint split into two different projects. Take a look at [Node.js API Starter Kit](https://github.com/kriasoft/nodejs-api-starter) as an example. You can either have front-end and API deployed under different domains (e.g. `www.example.com` + `api.example.com`) or, event better, deploy them under the same domain via a reverse proxy such as Nginx or IIS (`www.example.com` front-end + `www.example.com/graphql` data API).
- [x] Introduce `API_URL` environment variable, set it to `https://reactstarter.com` by default to make the site work with the existing GraphQL API.
- [x] Replace `core/fetch` with `core/createHttpClient`, instead of using `fetch` directly it's better to initialize a new instance of HTTP client for each website user, this unlocks caching, and server-side rendering with data fetched from external API.

If you prefer having data API in the same project with the front-end code, you're free to continue using that approach and it must be easier for you to resolve merge conflicts when you pull and merge latest changes from RSK back into your project.


langpavel:
This will make setup of RSK projects harder.

My suggestion is make this on `next` branch and make `next` branch default.

@koistya, @frenzzy what do you think?

buildbreakdo:
@koistya Node API Starter Kit is a valuable project no doubt, experience has been that large software projects especially with multiple teams are typically split like this.

Is the thought that most people are using the RSK to work on large software projects? 

If the issues are any guide, looks like people just starting out (small hobby projects to maybe medium sized projects) . Believe asking these groups -- especially the small hobby projects which probably comprise the majority -- to dive into yet another project if they want minimum functionality like a database is asking too much and would expect them to look at a different starter kit that includes a database.

Think about the development workflow, now you are running two projects at the same time? Now when a change occurs in one project you have to go to the other project to sync them, run them, debug them. Two sublime text editors open? Workflow somewhat makes sense for large organizations/teams/projects as teams have to negotiate, however do not think this is the right direction for a react _starter._ kit. If react has taught us anything has it not been that cohesion > coupling?

Honestly, think we should be more opinionated and provide better examples around those opinions. Last I checked we do not have a concrete usage example for the database, it's just there. Basic auth should be in the project.

Understand the drive to reduce complexity just do not think moving the complexity to another project achieves this goal -- actually think the opposite -- this is an increase in complexity.

kybarg:
Trying to start new project using `serverless` branch but get an error `Failed to load the news feed.`

buildbreakdo:
@kybarg Why use the serverless branch when you can use https://github.com/kriasoft/react-static-boilerplate? Not if the serverless branch is being kept up to date. Recommend using the react-static-boilerplate for your base project if you do not want a server.

kybarg:
@buildbreakdo cause I want to build `univeral` app.

langpavel:
This PR is quite outdated.
For me I must repeat that all-in-one is biggest advantage of this starter kit.
It enables things like direct server side GraphQL evaluation as I showed in [feature/apollo custom NetworkInterface implementation](https://github.com/kriasoft/react-starter-kit/tree/feature/apollo/src/core/createApolloClient)

koistya:
@Stupidism, everyone, what do you think about removing GraphQL/Auth part from RSK in favor of having data API (including authentication/authorization) in a stand alone project ([Node.js API Starter](https://github.com/kriasoft/nodejs-api-starter))?

Stupidism:
Great, the only thing I don't want in RSK is GraphQL now!
I think that's enough for the server of RSK to be able to handle proxy.

About jwt auth, in my own project, I used it to do SSR-auth. Otherwise, I didn't find a decent way to do SSR-auth without cookie.

koistya:
@Stupidism do you plan to use an external GraphQL server, or not to use it at all?

Stupidism:
I don't use GraphQL at all, it doesn't coordinate with our mirco-service architecture

langpavel:
@koistya It is much more easier for tooling, newcomers, everybody having all together.
It is easy to split code later when team decide. For me, working alone or in pair is really big plus having all together. Note that it is possible create extra entry point for GraphQL server, it is simple few lines in `tools/webpack.config.js` and lightweight express app..

pruhstal:
@koistya Is this an effort to move RSK to being independent of any database (per your commit @ #df3fe75ccf3952df356931da4eeec6a97770ef7c)?

Because if so, I'm in full support of that option.

How to get RSK to play nicely with node.js api starter kit would be much helpful in addition to merging that.

ulani:
@koistya  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1093, opened 2017-01-18
Update Webpack to v2.2; update other npm modules

koistya:
https://webpack.js.org/guides/migrating/


samkelleher:
@koistya in a few places there are still references to `config.debug` (for example in start.js) however the debug attribute from the config was removed, so now this condition never meets. The `LoaderOptionsPlugin` was added, but only presents to plugins and not something importing the config file.

Because debug is always false the Hot Module Replacement (HMR) + React Hot Reload are never enabled.

helly0d:
According to [migration guide](https://webpack.js.org/guides/migrating/#dedupeplugin-has-been-removed) the `DedupePlugin` has been removed. I think it's safe to remove it from the config as well.

sashakru:
Do you have any updates about when this pull request will be merged to master?

frenzzy:
Thanks guys for comments, updated and merged in #1108

PR #1113, opened 2017-02-02
Adds error handler for errors that might happen after rendering.

phpeek:
Currently when there's an error in e.g. `onClick` handler, error is being shown only in browser console.

It might flew unnoticed by developer.

After this change it should show up in full screen for development mode.

@nupurkapoor, @GalGavu, @noahehall and me are using it for our project and it works just fine so we've decided it might be worth to introduce it to Your codebase.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -176,6 +176,18 @@ async function onLocationChange(location) {
 history.listen(onLocationChange);
 onLocationChange(currentLocation);
 
+// Handle errors that might happen after rendering
+// Display the error in full-screen for development mode
+// Fallback to console.error in production mode
+window.addEventListener('error', (e) => {
+  console.error(e.error); // eslint-disable-line no-console
+
+  if (process.env.NODE_ENV !== 'production') {
+    document.title = `Error: ${e.error.message}`;
+    ReactDOM.render(<ErrorReporter error={e.error} />, container);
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -176,6 +176,18 @@ async function onLocationChange(location) {
 history.listen(onLocationChange);
 onLocationChange(currentLocation);
 
+// Handle errors that might happen after rendering
+// Display the error in full-screen for development mode
+// Fallback to console.error in production mode
+window.addEventListener('error', (e) => {
+  console.error(e.error); // eslint-disable-line no-console
```
frenzzy: Why is it necessary to attach such event listener in release mode?

phpeek: You're right. There's absolutely no sense doing that. Let me update PR.

phpeek:
Strange, build failed due to:

```
curl: (22) The requested URL returned error: 404
> Failed to download https://yarnpkg.com/latest.tar.gz.
````

Let me amend my commit and force update a branch, so that buid will kick in again.

phpeek:
Looks like something is up with yarn website. It 404s all over the place. Will wait till it's back to a normal state.

langpavel:
@koistya, your turn ;-)

PR #1108, opened 2017-01-31
Update Webpack to v2.2.1; update other npm modules

frenzzy:
continues #1093
also fixes #1047 #1021 #954

PR #1115, opened 2017-02-05
Use autoprefixer for babel: compile less, polyfills only for supported environments

frenzzy:
- Integrate autoprefixer for Babel that can automatically determine the Babel plugins to compile only for your supported environments: [`babel-preset-env`](https://github.com/babel/babel-preset-env)
- Move `browserslist` to `package.json` for `autprefixer`, `stylelint` and `babel-preset-env`
- Replace `babel-preset-stage-0` with `babel-preset-stage-2` to avoid using [unstable features](https://babeljs.io/docs/plugins/#presets-stage-x-experimental-presets-) in production
- Remove `babel-plugin-transform-runtime` dependency in favor of `babel-polyfills`
- Remove `extend` dependency

### DIFF
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -9,8 +9,8 @@
 
 import path from 'path';
 import webpack from 'webpack';
-import extend from 'extend';
 import AssetsPlugin from 'assets-webpack-plugin';
+import pkg from '../package.json';
 
 const isDebug = !process.argv.includes('--release');
 const isVerbose = process.argv.includes('--verbose');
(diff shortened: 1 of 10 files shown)


tools/webpack.config.js
```diff
@@ -155,17 +106,68 @@ const config = {
...
+                browsers: [ // keep in sync with package.json/browserslist
+                  '>1%',
+                  'last 4 versions',
+                  'Firefox ESR',
+                  'not ie < 9',
+                ],
+              },
+              modules: false,
+              useBuiltIns: true,
+              debug: false,
```
phpeek: Can value of `debug` come from `isDebug` ?

frenzzy: I did not use `isVerbose` here because if I set `debug: true` for both client and server configs I see debug log only for first of them. So, I think maybe it is better to change the option only when you need it.

langpavel:
@frenzzy 
I have issue with this:
I have circular imports in my GraphQL tree. After merging this commit into my tree, development server didn't start correctly, because I circularly imported references was not updated in place, instead they stayed `undefined` even after they should be defined. May be badly implemented hoisting of this..

frenzzy:
@langpavel,
Babel don't touch Imports and exports after [migration to webpack 2](https://github.com/kriasoft/react-starter-kit/pull/1108). In this PR we just stop using a lot of plugins from [babel-preset-latest](https://babeljs.io/docs/plugins/preset-latest/) for server side bundle, because Node.js supports many things out of the box (assumes that you use node >= 6.5).

Function declarations in babel are hoisted: https://github.com/babel/babel-preset-env/issues/112#issuecomment-277467687
But probably there are a real problem in your codebase with native behavor.

You still can migrate to actual changes and continue using `latest` preset for server side only [here](https://github.com/kriasoft/react-starter-kit/blob/52899e83f83255b8f30968396f3300e8c1b6c3fd/tools/webpack.config.js#L261-L268) until you figure out the problem in your codebase.
```js
['latest', { es2015: { modules: false } }],
```

langpavel:
Hmm, I'm sure problem is in webpack, not sure.., may be because of `export default` construct...
I found this issue, https://github.com/webpack/webpack/issues/1788

frenzzy:
This PR disables webpack 2 tree-shaking, maybe it is related (reenabled back in 7bbe94c2508203b7a941120ea6a63d9fcf2fc837)

PR #1131, opened 2017-02-11
add a dynamic navigation by graphql query me

ClemensSahs:


langpavel:
@ClemensSahs You can use router itself with all advantages to create custom object and then pass it into navigation component as a `prop`

ClemensSahs:
@langpavel sound great, after a short doc scanning form "universal-router" I'm not sure that look great, but maybe I'm wrong.

In order to have a man navigation in the header another one in the footer and perhaps a content left.
That will make thinks complicated, right?
I really like the idea, to centralize the navigation data.

langpavel:
Huh, sorry @ClemensSahs, I'm not sure if I understood you well, but centralizing navigation in router is definitely better idea then have separate decision tree in component or it's helper, I'm right?

You can introduce new property returned from each route and possibly merge it's data from higher routes.. Yes man?

ClemensSahs:
Yes its the best Idea, to centralizing the navigation. But in my mind the router looks not the right place, perhaps I'm wrong. 

Currently I don't have the full overview from the react ecosystem.
What I searching for is more like a navigation-route-mapper. 

So the component call the only the mapper->get('head-navi') or mapper->get('content-navi')

Navigation case:

- Head Navi
  - left region
    - Home 
    - about
    - Contact
  - right region
    - login and register (if not login)
    - dropdown (if login)
      - dashboard
      - logout
- content Navigation (if route == settings)
  - Profile
  - Billing
  - Notifications
- Footer
  - left Region
    - terms
    - Privacy
  - right region
    - Contact
    - Blog
    - About

langpavel:
I'm going to close this, but I'm really interested if you find this helpful for your use case. Please, write your observations here. Anyway, Thanks!

PR #1133, opened 2017-02-16
Add VS Code settings

langpavel:
[Visual Studio Code](https://code.visualstudio.com/) default settings

This is better than adding `.vscode/` to `.gitignore`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "vsicons.presets.angular": false
+}


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "vsicons.presets.angular": false
```
frenzzy: This option is not a part of standard VS Code configuration. In general I think it is not a bad idea to keep some editor configuration in the repo, but do not understand why we need this option or something which will duplicate or override `.editorconfig`.

Take a look at [aspnet-starter-kit/.vscode](https://github.com/kriasoft/aspnet-starter-kit/tree/master/.vscode) folder.

langpavel: @frenzzy 
Ok, this is first shot, but definitely, if you have `vsicons` extension installed, it will ask you for restart your editor each time you open non-angular project.

`vsicons` is heavily spread extension so this definitely helps everyone who have this installed.

Another think for example is to disable internal `js` validator, because of `flow`. But at first time you should have `flow` extension installed in VSCode, so I didn't include this in yet, because I'm new to vscode (but I'm really excited :-) ) and I don't know how to suggest recommended packages to VSCode...

After this file will be included in repo it is possible that we will receive more PR for `vscode` editor, so I think this first step will kick others to send more PRs.

frenzzy: I have the same [vscode-icons](https://marketplace.visualstudio.com/items?itemName=robertohuertasm.vscode-icons) plugin installed but it doesn't ask me to restart the editor for any of my RSK based project.

langpavel: Hmm.. @frenzzy have you some user Preferences/settings config?

frenzzy: Have no options related to this plugin.
Workspace Settings: default
User Settings: default extended with
```js
{
  // The number of spaces a tab is equal to. This setting is overriden based on the file contents when `editor.detectIndentation` is on.
  "editor.tabSize": 2,

  // Controls auto save of dirty files. Accepted values:  "off", "afterDelay", "onFocusChange" (editor loses focus), "onWindowChange" (window loses focus). If set to "afterDelay", you can configure the delay in "files.autoSaveDelay".
  "files.autoSave": "onFocusChange",

  // Controls how the editor should render whitespace characters, posibilties are 'none', 'boundary', and 'all'. The 'boundary' option does not render single spaces between words.
  "editor.renderWhitespace": "boundary",

  // When enabled, will trim trailing whitespace when you save a file.
  "files.trimTrailingWhitespace": true,

  // Insert a final newline when saving the document.
  "files.insertFinalNewline": true,

  // Settings for PostCSS Sorting plugin
  "postcssSorting": {
    "preserve-empty-lines-between-children-rules": true,
  },
}
```

langpavel: Ok, @frenzzy, have you an idea how to set defaults? I thins we can merge some of mine and some of yours which is already dictated with eslint rules. so this?

```json
{
  // The number of spaces a tab is equal to. This setting is overriden based on the file contents when `editor.detectIndentation` is on.
  "editor.tabSize": 2,

  // When enabled, will trim trailing whitespace when you save a file.
  "files.trimTrailingWhitespace": true,

  // Insert a final newline when saving the document.
  "files.insertFinalNewline": true,

  // Settings for PostCSS Sorting plugin
  "postcssSorting": {
    "preserve-empty-lines-between-children-rules": true,
  },
}
```
`"vsicons.presets.angular"` is no longer needed after https://github.com/vscode-icons/vscode-icons/pull/781

frenzzy: All this options will be overwritten by [.editorconfig](http://editorconfig.org/) file, so I think that is not necessary.

Except postcssSorting which is not longer working in VS Code and maybe it is better to configure [stylelint-order](https://github.com/hudochenkov/stylelint-order) [plugin](https://stylelint.io/user-guide/plugins/) for [stylelint](https://stylelint.io/) then with pre-commit hook.

PR #1134, opened 2017-02-16
Webpack bundle analyzer

langpavel:
* New `--analyse` flag for webpack 
 * New task `build:analyse` — alias for `yarn run build -- --release --analyse`

Thank you @frenzzy for great suggestion!

image


tools/webpack.config.js
```diff
@@ -216,6 +232,31 @@ const clientConfig = {
...
+      // Name of Webpack Stats JSON file that will be generated if `generateStatsFile` is `true`.
+      // Relative to bundles output directory.
+      statsFilename: path.resolve(__dirname, '../stats.json'),
+      // Options for `stats.toJson()` method.
+      // You can exclude sources of your modules from stats file with `source: false` option.
+      // See more options here: https://github.com/webpack/webpack/blob/webpack-1/lib/Stats.js#L21
+      statsOptions: null,
+      // Log level. Can be 'info', 'warn', 'error' or 'silent'.
+      logLevel: 'info',
+    }),
```
frenzzy: Why don't you want to use default host and port?
Why you still need to generate `stats.json` file?
Maybe do not specify options which already are defaults to reduce the overall size of the config and keep it as simple as possible?

Maybe it would be nice to use `static` mode for `yarn run build -- --analyze`
and `server` mode for `yarn start -- --analyze`, what do you think?

langpavel: > Why don't you want to use default host and port?

What is default is dependent. Default is 8888, ok, but why don't keep ports together? RSK is using 3000-3002 today..

> Why you still need to generate stats.json file?

Ok, it isn't generated for development and in fact it is only generated when you building for production.. not bad choice by my opinion

> Maybe do not specify options which already are defaults to reduce the overall size of the config and keep it as simple as possible?

It is better to have comments and explicit defaults.. Few bytes as comments are better than ten minutes of searching and reading docs..

> Maybe it would be nice to use static mode for `yarn run build -- --analyze` and server mode for `yarn start -- --analyze`, what do you think?

`yarn start -- --analyze` is not so useful if you are looking for possible optimizations in production bundle... but if you want see last production build, you can look at `webpack-bundle-analyzer stats.json` directly...

frenzzy: This is very opinionated but I think It is much easier to start working with new boilerplate when it has as less configuration options as possible and overrides defaults for other libs only when that really matters. Also users which familiar with lib may expect default values too. Btw, Browsersync ports were changed to defaults in 2da58da1b252e1a6502319cc414d556453b96450.

If we will start adding all possible options with default values to each plugin or loader to webpack config then the already difficult-to-understand configuration will become enormous and turn into a mess. Also it will be impossible to understand what exactly has changed in comparison with the initial values. That's why we just use a small description with a link to the readme near each feature and specify only important options.

I still do not understand why you need both generate `stats.json` and launch analyzer together. I don't think stats file will be useful because the data is actually for specific build and every time when you will want to analyze them you will not be sure whether these are relevant or not. Stats file exactly is not needed in the project's root directory and useful only if you will put it to the build folder and will deploy together with production files to be able to analyze production bundle at any time. Also generating of `stats.json` in the root folder is useless if you prepare production build in the cloud (CI, Docker, etc.).

langpavel: Ok @frenzzy, you are right.. but I'm in hurry now

PR #1141, opened 2017-02-20
Remove browser-sync

apsavin:
This PR removes browser-sync dependancy (and functionality).

Now console ouput in the end of `start` job looks like this:


langpavel:
Why?

apsavin:
@langpavel 

Because this dependency is huge and (in my expirience, which is not relevant, of course) most of the devs wan't use it anyway. 

Feel free to just close the PR.

langpavel:
@apsavin It is possible to make it *opt-out*. Do you want try this on current master?

This means. If you don't want use it, you can disable it on command line, like `yarn start -- --no-browser-sync` or switch this by environment variable, like `BROWSER_SYNC=disable yarn start`.. This will need only few conditionals in `tools/` but it is true I'm using browsersync only when I tuning styles on mobile and PC, so relatively rarely..

apsavin:
@langpavel well, it is possible, but users will download it then.

PR #1145, opened 2017-02-22
Add ability to import global styles

zazido:


frenzzy:
Usually you need `:global` scope only for third-party css files which you cannot rename because they are in `node_modules` directory.

If you need to completely disable `:local` scope for css files in the project, just set `modules: false` for css-loader.

Alternative solution in #1153

PR #1161, opened 2017-03-07
Split webpack config to separate files

langpavel:
It makes more sense to keep webpack configuration in separate files.


frenzzy:
I think we should not split part of configuration (common + client) to different files because it is a part of single config, it is easy to see all possible options in one place. The same for server config which is just a bit modified version of client config. Also we should not bundle server-side code (api backend) by webpack in the future.

langpavel:
Thanks @frenzzy for your opinion, but I'm on opposite:
> I think we should not split part of configuration (common + client) to different files because it is a part of single config

I'm not sure, it is easy to modify different part of config.

> Also we should not bundle server-side code (api backend) by webpack in the future.

Yeah, but not now, because of different `require` file types. Also note, that [`require.extensions`](https://nodejs.org/api/globals.html#globals_require_extensions) is deprecated, so this should be no go for future:
> > Deprecated In the past, this list has been used to load non-JavaScript modules into Node.js by compiling them on-demand. However, in practice, there are much better ways to do this, such as loading modules via some other Node.js program, or compiling them to JavaScript ahead of time.
> > 
> > Since the module system is locked, this feature will probably never go away. However, it may have subtle bugs and complexities that are best left untouched.
> >
> > Note that the number of file system operations that the module system has to perform in order to resolve a require(...) statement to a filename scales linearly with the number of registered extensions.
> >
> > In other words, adding extensions slows down the module loader and should be discouraged.

And at last. I need CLI utility with shared code, so easiest way to go is to create a new webpack configuration based on servers one OR extend server configuration.
I don't see reason why you should not be able to operate by webpack command line on different parts of application separately if you wish. Now it will be possible.

frenzzy:
Usually you do not need to change any webpack config options, because everything is pre configured out of the box. But when you open webpack config file first time it's much easier when **different keys of the same configuration object** defined in the same place (or file at least).

For server side we can just reuse result of client side webpak compilation process. For example import will stay the same: `import s from './Header.css';` just no more css in the file:
```js
// Header.css
module.exports = {
  root: 'qwert',
  logo: 'werty',
}
module.exports.getCss = () => '.qwert{position:relative}.werty{color:red}';
```

If you want to reuse server side config you can just do like this:
```js
let [, serverConfig] = require('./webpack.config.js');
// reuse serverConfig only
```
Do not understand how the config that separated into different files related to webpack's command line or new possibilities.

langpavel:
> Do not understand how the config that separated into different files related to webpack's command line or new possibilities.

Because then you can run webpack for server separately to collect stats, for example...

koistya:
I think that in the next few days we will remove Webpack dependency for the server-side code, so that it will be compiled directly by Babel, similarly to [Node.js API Starter](https://github.com/kriasoft/nodejs-api-starter). `webpack.config.js` will become shorter and fit within 1-2 screens. If so, splitting it would make things harder to find and edit.

langpavel:
Ok, @koistya 
1. Will transition to babel without webpack for server can be beneficial in some way?
2. If you split it, what things will be harder to find and edit?

frenzzy:
@langpavel 
> 1. Will transition to babel without webpack for server can be beneficial in some way?

Webpack is a bundler, but server side code should not be bundled (similarly to Node.js API Starter leads to the fastest compilation and hot reload)

> 2. If you split it, what things will be harder to find and edit?

All keys of the same configuration object should be defined in the same file. Easier to search for keys.

> 3. It you remove webpack from server processing, things become very different, so splitting this makes more sense now then ever!

Code splitting is client-side only feature, all server side files always available in the file system.

> 4. It you remove webpack from server processing then you must replace `markdown-loader`, `raw-loader` and more.. I cannot imagine how you replace `isomorphic-style-loader` *in nice and effective way* easily.

We will just reuse result of the loaders for server side, it makes no sense to run the loaders twice  for the same files.

> Because then you can run webpack for server separately to collect stats, for example...

Analyzer tools allow you to see stats separately for each file, bundle, entry, module, etc. So, you do not need to do with configs anything.

If you still want to have separate configs you can do like this in your code base:
```js
// tools/webpack.client.js
const [clientConfig] = require('./webpack.config.js');
export default clientConfig;
```
```js
// tools/webpack.server.js
const [, serverConfig] = require('./webpack.config.js');
export default serverConfig;
```

PR #1169, opened 2017-03-12
env/add-idea-conf-to-git-ignore

flexpert:
Adde `.idea` folder to `.gitignore` for all Idea based IDE users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ database.sqlite
 node_modules
 ncp-debug.log
 npm-debug.log
+.idea


frenzzy:
IDE settings are usually configured in a global .gitignore as described here:

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

frenzzy:
See for example [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)

PR #1174, opened 2017-03-14
ApolloClient/createNetworkInterface credentials

mika75:
To pass basic auth

createApolloClient.client.js

```
const client = new ApolloClient({
  networkInterface: createNetworkInterface({
    uri: '/graphql',
    opts: {
	    credentials: 'include',
    },
  }),
  queryDeduplication: true,
  reduxRootSelector: state => state.apollo,
});

export default function createApolloClient() {
  return client;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ branches that you can use either as a reference or merge into your project:
     Simplest possible integration of [react-bootstrap](https://react-bootstrap.github.io/)
     by [Pavel Lang](https://github.com/langpavel) (based on `master`)
 
+You can see status of most reasonable merge combination as [PRs labeled as `TRACKING`](https://github.com/kriasoft/react-starter-kit/labels/TRACKING)
+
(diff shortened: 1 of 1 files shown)


langpavel:
@mika75 Thanks! But is commit reflecting what you propose? :-)

mika75:
@langpavel Thank !

GGAlanSmithee:
BTW, here are the `credentials` options;

* omit: Never send cookies.
* same-origin: Only send cookies if the URL is on the same origin as the calling script.
* include: Always send cookies, even for cross-origin calls.

[source](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)

If possible, `same-origin` should probably be used.

langpavel:
I will keep include because it's same in this case. And of you move GraphQL endpoint to different domain, you will have hard times of you overlook this :-)

PR #1183, opened 2017-03-19
Implement email login flow

jamalx31:
@langpavel 

Need help to move forward with this:
1- not sure how to connect this with Passport
2- if I added username/password in the right place
3- if bcrypt is needed in the first place

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "dependencies": {
     "apollo-client": "^1.0.0-rc.2",
     "babel-polyfill": "^6.22.0",
+    "bcrypt": "^1.0.2",
     "bluebird": "^3.4.7",
     "body-parser": "^1.16.0",
     "classnames": "^2.2.5",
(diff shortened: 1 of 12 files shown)


langpavel:
Hi, personally I don't use passport. I created my own password login flow in private project but I'm not sure if way I do it is right. But it works for me and I can describe how it works.

At first, you need endpoint for login. I integrated this into GraphQL as mutation but I see more often now that doing this as simple REST can be less hacky, simpler and maintainable.

Then if request succeed you can set HTTP-only cookie. You can do this in GraphQL too but it feels .. well.. not so nice..

I'm using JWT in cookie which have ACL array of strings. 

If user is authenticated, JWT is parsed on server and ACL is serialized to redux store. Then I have ACL React component which provide ACL instance to it's child. So everything can be handled declaratively

piglovesyou:
As one of sample cases, my project uses `passport-local` for local DB authentication, which is very common way for local DB auth in Node.js I could say and is very good fit to RSK-base project, since `passport-facebook` is already in. I think it's a good idea to integrate `passport-local` impl in `master` to make email-auth work out of the box.

PR #1184, opened 2017-03-19
Fix jwt error signing recursive json when logging in from second browser

velocityboy:
When signing in from a second browser, jwt signs data that comes from sequelize rather than the auth provider. By default this blob contains a bunch of table metadata and some circular references, so the sign fails. This diff asks sequelize to give back just the table data and strip out all the problematic metadata.

### DIFF
--- a/src/core/passport.js
+++ b/src/core/passport.js
@@ -82,7 +82,8 @@ passport.use(new FacebookStrategy({
         ],
       });
       if (users.length) {
-        done(null, users[0]);
+        const user = users[0].get({ plain: true });
+        done(null, user);
       } else {
         let user = await User.findOne({ where: { email: profile._json.email } });
         if (user) {


langpavel:
***logging in from second browser*** what?? It taste same as different user, right?

velocityboy:
Log in, for example, to an RSK site via Facebook in Chrome. Then, log in as the Facebook user on Safari. Without the fix, the second login will fail due to the extra data in the the cached login in the db.

langpavel:
Ok, but it's yous server, not browser which is affected. BTW this is purely `passport` thing... This is reason... I need know whats happening when user register or login. There are most attack vectors and analysis leakage when you don't understand this flow..

If you want use 3rd party authentication it does not means you should not care about whats happening in backend!

velocityboy:
Right, I get it's on the server, but failing to be able to pull the user information from the database causes the flow the fail at that point and the error is propagated back to the browser.

langpavel:
> causes the flow the fail at that point and the error is propagated back to the browser

Is this really the case? Or it is happening in opposite direction?

velocityboy:
Here's what happens in all the gory detail, using Facebook as an example:

1. User initiates login in the browser and the browser navigates to /facebook/login
2. Server calls passport.authenticate. The Facebook strategy begins an OAuth request to Facebook, potentially returning a browser redirect for the user to agree to the auth on the Facebook side
3. Facebook redirects the browser to the completion url /facebook/login/return back on our site
4. On the server, passport.authenticate is called a second time to make sense of the data returned by Facebook
5. The passport callback is called. In the case we're talking about, this isn't the first time the user has logged into our site, so there's already a User record in the sequelize store for the user.
6. The code in the RSK passport wrapper tries to pull the existing user data (based on the Facebook identity returned in the previous step) out of the database and call the passport done() callback on that user's record.
7. Sequelize returns the user record, but it has a bunch of metadata in in which happens to contain a circular reference. Passport tries to call JSON.stringify on the object, and an exception gets thrown.
8. Login fails because of the exception.

jorrit:
This should be merged. Another solution is to add `raw: true` to the sequelize query.

@velocityboy : can you resolve the conflicts?

langpavel:
@koistya @frenzzy I don't see anything dangerous here. Can be merged?

PR #1188, opened 2017-03-21
Add CSS modules :local pseudo selector

Deepblue129:
Along with the :global selector, CSS modules supports :local as an integral part of their library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,7 +172,8 @@
         true,
         {
           "ignorePseudoClasses": [
-            "global"
+            "global",
+            "local"
           ]
         }
       ]


langpavel:
Is this really needed or you thing this can add *points* to code clarity?

Deepblue129:
This is needed for working effectively with code modules.

For example, in my project, I have a ".grid" selector. It helps me build grids across many different React Modules. I would mark the ".grid" selector ":global" while it's children are local scope.  To shift back to local scope, I must use the ":local" selector.

Snippet from my code where I found I must use ":local"
```
    > :global(.grid) {
      align-items: center;
      flex-direction: column;
      justify-content: center;
      text-align: center;
      height: 100%;

      :local {
        h2 {
          margin-bottom: 1rem;

          @media (min-width: 550px) {
            margin-bottom: 1.5rem;
          }
        }
      }
    }
```

langpavel:
Huh, I'm not so good in this. @frenzzy can you help, please?

frenzzy:
CSS Modules local by default in RSK, you don't need to use `:local`:
```css
:global(.outer) {
  color: green;
  .inner { color: red }
}

/* output */
.outer { color: green; }
.outer .file-inner-qwert { color: red; }
```

```css
:global .outer {
  color: green;
  .inner { color: red }
}
/* or */
:global {
  .outer {
    color: green;
    .inner { color: red }
  }
}

/* output */
.outer { color: green; }
.outer .inner { color: red; }
```
also it would be nice to not mix local and global classes together:
```css
:global(.outer) { color: green; }
.inner { color: red; }
```

Deepblue129:
What is the main contention against this pull request? 

- :local is a core feature of css-modules out of 2.
- CSS-Modules does not cite splitting local/global selectors as a best practice.  @frenzzy
- CSS-Modules documents relies on the use of :local. Here is an example on their front page
`.localA :global .global-b .global-c :local(.localD.localE) .global-d`.
- This change is minuscule with no adverse effects on the users. It has only the benefit of properly supporting a third party library.

langpavel:
Landed in c3f0c85e387a2df8579701ab871842386325ab97. Thank you

Issue #1195, opened 2017-03-27
React-select on safari mobile does not work 


martinratinaud:
Hi guys, I forked the project and made a branch to illustrate this problem.

https://github.com/martinratinaud/react-starter-kit/tree/fix/react-select

Launch `npm start` and go to http://localhost:3000 from a mobile (not the chrome simulator but a real device or the iOS Simulator)

When you try to select `Two`, nothing happens even though it works well on a desktop browser.

In the meantime, the official page of react-select deos work on safari mobile http://jedwatson.github.io/react-select/

I tested by disabling SSR but it does not work either.

I think the problem comes from the webpack config but I could not figure this out.

If somebody could help me on this, that would be awesome.

PS: I experienced the same with react-dropzone but hopefully fixing this will fix it too

frenzzy:
Maybe it requires some polyfills which should be executed before the lib?
Do you have [this change](https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L161-L163) in your RSK version?

martinratinaud:
Hi, 

I have the change you mention and I don't have any error, it is just that I can't select the `two` entry as the `logChange` function is never triggered on safari mobile

phpeek:
@martinratinaud What happens if You use port 3001 instead of 3000 ?

Jaikant:
I had a similar issue with react-select, using port 3001 was a workaround which worked for me.

martinratinaud:
Thanks for the insights

Actually, I tried it on production env and I get the same problem.
I also disabled SSR and I still have the problem (see the updated repo https://github.com/martinratinaud/react-starter-kit/tree/fix/react-select)

Finally if I make a long click on a select option it will select it but if I just tap/click it won't (in safari mobile again)

Again, react-select is working fine outside of react-starter-kit.

Thnaks for your help

Jaikant:
@martinratinaud Trying to recall what else I had to change, one other thing which I had to modify for react-select and mobile was I moved to 'react-fastclick'

In client.js

// Commented out the existing Fastclick and used react-fastclick instead.
//import FastClick from 'fastclick';
import initReactFastclick from 'react-fastclick';

//FastClick.attach(document.body);
initReactFastclick();

martinratinaud:
Thanks a lot @Jaikant it works well if I replace `fastclick` with `react-fastclick`. I'll submit a PR for this.

Cheers

PR #1198, opened 2017-03-29
use react-fastclick instead of fastclick

martinratinaud:
In order to fix the bug https://github.com/kriasoft/react-starter-kit/issues/1195. 

Thanks to the help of the community, I found out using fastclick was causing problems when trying to use `react-select` on a safari mobile. I made a demo to illustrate this problem/solution here

https://github.com/martinratinaud/react-starter-kit/tree/fix/react-select-demo

with `fastclick` options are not selectable on safari mobile
with `react-fastclick` it works

Thanks


package.json
```diff
@@ -35,6 +35,7 @@
     "query-string": "^4.3.1",
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
+    "react-fastclick": "^3.0.1",
```
langpavel: You add new dependency but you not removed old one.

martinratinaud:
Ooops, sorry about that, here you go I updated the PR

frenzzy:
Considering to [original fastclick docs](https://github.com/ftlabs/fastclick#when-it-isnt-needed), maybe we should remove fastclick at all?

martinratinaud:
@frenzzy I don't get it. What part of the doc are you refering to exactly ?

Thanks

langpavel:
@martinratinaud Slightly extracted what @frenzzy refs to:

## When it isn't needed
 - Desktop browsers.
 - Chrome 32+ on Android with `width=device-width` in the viewport meta tag
 - Chrome (all) on Android with `user-scalable=no`
 - For IE11+, you can use `touch-action: manipulation;` to disable double-tap-to-zoom on certain elements (like links and buttons). For IE10 use `-ms-touch-action: manipulation`

## When it is needed?
- [ ] Apple software/hardware? Don't know
- [ ] Mobile Firefox? Don't know

`react-fastclick` patch React internals in way I really dislike..

martinratinaud:
I tried to remove react-fastclick and could **not** make it work on my iPhone 5s in both safari and chrome so I guess it's not ready yet.

shishirarora3:
Any resolution yet? Should we remove fast-click completely?

jorrit:
@martinratinaud what do you mean? What doesn't work and in which situation?

Bamieh:
talking from personal experience, fastclick is useless with react, we ended up writing a custom solution at the component level, wrapping any clickable we want, something like `react-tappable`.

frenzzy:
Thanks for discussion, [fastclick](https://github.com/ftlabs/fastclick) dependency was removed from RSK in #1305

PR #1208, opened 2017-04-05
Fix Request-type object passing to localFetch

dagen-niger:
Hello

It is impossible to import `Request` from `node-fetch` and pass this request as parameter to localFetch from `fetch.server.js`, however node-fetch support this since Feb 2016. This PR will fix this SSR issue taking into account relative urls.

So now able to write smth like this:
```javascript
import fetch, { Request, Headers } from '../core/fetch'

function prepare(method, path, data) {
  // prepare path and options here
  return new Request(path, options)
}

async function execute(req) {
  try {
    const response = await fetch(req)
    // ...
  }
  // ...
}

```

### DIFF
--- a/src/core/fetch/fetch.server.js
+++ b/src/core/fetch/fetch.server.js
@@ -14,6 +14,10 @@ import { host } from '../../config';
 fetch.Promise = Promise;
 Response.Promise = Promise;
 
+/**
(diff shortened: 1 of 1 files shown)


langpavel:
This looks quite outdated.

PR #1210, opened 2017-04-05
Comments and optimize eslintConfig

Stupidism:


package.json
```diff
@@ -134,6 +134,25 @@
       }
     }
   },
+  "//eslintConfig": {
```
langpavel: I don't like this style of comments in JSON, but right now, they are only possible..

But when you nest keys under, it goes so unclear.. so if you want comment nested keys, do it in same way, prescribe any commented out key by double-slash but try to not nest it...

package.json
```diff
@@ -134,6 +134,25 @@
...
+    "rules": {
```
langpavel: Is this valid rule? no. because parent is comment only, so, make this comment. This rule is 4 lines only so it can be clear for everyone that it is commented out...

package.json
```diff
@@ -150,8 +169,17 @@
       "browser": true
     },
     "rules": {
-      "import/extensions": "off",
       "import/no-extraneous-dependencies": "off",
+      "no-console": [
+        "error",
```
langpavel: why? "warning" or "warn" (I'm not sure) is good level for me. We have many disabled `console` in code (Yeah, my branches especially..)

src/client.js
```diff
@@ -10,7 +10,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import FastClick from 'fastclick';
-import UniversalRouter from 'universal-router';
+import UniversalRouter from 'universal-router/main';
```
langpavel: huh? I don't know if this is the recommended one @koistya, @frenzzy?

package.json
```diff
@@ -150,8 +169,17 @@
...
+        {
+          "allow": [
+            "warn",
+            "error",
+            "info"
```
langpavel: others are error, but you allow `info`? Does not make sense for me.. Will be better to enclose console messaging with `!__DEV__` conditions and explicitly mark them with eslint hints.

src/client.js
```diff
@@ -134,7 +134,7 @@ async function onLocationChange(location, action) {
       throw error;
     }
 
-    console.error(error); // eslint-disable-line no-console
+    console.error(error);
```
langpavel: ^^^ as I said, enclose in `if(!__DEV__)`, keeping eslint comment hint here

src/server.js
```diff
@@ -16,7 +16,7 @@ import expressGraphQL from 'express-graphql';
 import jwt from 'jsonwebtoken';
 import React from 'react';
 import ReactDOM from 'react-dom/server';
-import UniversalRouter from 'universal-router';
+import UniversalRouter from 'universal-router/main';
```
langpavel: Same as above

src/server.js
```diff
@@ -139,7 +139,7 @@ pe.skipNodeFiles();
 pe.skipPackage('express');
 
 app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
-  console.log(pe.render(err)); // eslint-disable-line no-console
+  console.error(pe.render(err));
```
langpavel: this should be keeped for production too, so eslint hint is on the place here..

src/server.js
```diff
@@ -156,10 +156,8 @@ app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
 //
 // Launch the server
 // -----------------------------------------------------------------------------
-/* eslint-disable no-console */
 models.sync().catch(err => console.error(err.stack)).then(() => {
   app.listen(port, () => {
-    console.log(`The server is running at http://localhost:${port}/`);
+    console.info(`The server is running at http://localhost:${port}/`);
```
langpavel: nice, server only, but severity is more important then info or log...
In node context it is the same.. so I'm not sure if this change is needed...
This is real one for dedicated logger ;-) but out of scope of this kit

Stupidism:
Sorry, I see you are treating console methods from level perspective, like `trace` < `debug` < `info` < `warn` < `error` ?
In my opinion and habit, `console.log` is used to print sth in develop process because it's the shortest and has no level. Sometimes, I would forgot to delete these lines and make commits dirty. So, if I do want to print sth in dev or prod env, I would use `console.error` `console.warn`... instead.

src/server.js
Stupidism: This would cause a problem, I would revert this

src/client.js
Stupidism: This would cause a problem, I would revert this

package.json
Stupidism: Do you mean make it `//rules` and `//import/extensions`?

package.json
Stupidism: Because package.json only support root comments, I think commenting root keys is enough to represent this is whole comments.  
I made it nested is to in case more comments in the future.
```
  "//eslintConfig": {
    "plugins": [
      "css-modules is for blablabla..."
    ],
    "rules": {
      "import/extensions": [
        "`js` and `jsx` are common extensions",
        "`mjs` is for `universal-router` only, for now"
      ],
    }
  },
```
is much clearer and more natural than 

```
  "//eslintConfig": [
    "plugins css-modules is for blablabla...",
    "rules import/extensions:  `js` and `jsx` are common extensions `mjs` is for `universal-router` only, for now"
  },
```

Stupidism:
I think you can re-review now. @langpavel 
About `console.log` @koistya also agreed with a thumb.
From code aspect, `console.info` is the same as `console.log`, so `info` can replace `log` where it's necessary.
From habit aspect, `console.log` is very easy to be abused, so we should forbit it.

package.json
langpavel: Ok, this cannot be overlooked :+1:

package.json
langpavel: Only `console.log` will be reported by lint, ok :+1:

src/client.js
langpavel: Ok, this is useful in production..

langpavel:
@Stupidism ping me after five days, I will merge if @koistya will not :-)

koistya:
What do you guys think about moving ESLint config back into `.eslintrc.js` file? Where you could comment the code and it might be easier for new users find ESLint config this way.

langpavel:
Hi @koistya, https://github.com/Stupidism/react-starter-kit/pull/2 definitely looks better.. So, no drawbacks only benefits?

koistya:
@langpavel not that I'm aware of

Stupidism:
Cool, I was thinking about the same thing. @koistya @langpavel

PR #1212, opened 2017-04-05
Fix date parsing with momentjs

### LINKED ISSUE #1211: feature/react-intl issue with Safari on OSX
Error formatting relative time.
getting enormous number of errors in console

node 6.9.2
yarn 0.19.1

```
[React Intl] Error formatting relative time.
```

image

zirho:
fix #1211

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "intl-locales-supported": "^1.0.0",
     "isomorphic-style-loader": "^1.1.0",
     "jsonwebtoken": "^7.2.1",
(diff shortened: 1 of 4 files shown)


langpavel:
Sorry, this is not fix, this is bad workaround which makes use of `<FormattedRelative>` component useless. You need polyfill https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat

zirho:
Hi, I see that.

PR #1213, opened 2017-04-06
feat(test-coverage): add coverage script

Stupidism:


Stupidism:
A wierd problem found:
https://github.com/istanbuljs/istanbuljs/issues/28
https://github.com/yarnpkg/yarn/issues/3056

However, it should not matter much. Just leave links here in case someone encounters this problem in the future.

Stupidism:
I fixed the diff again. Please review and merge. @koistya  @frenzzy

Stupidism:
Now, jsx is also covered well after https://github.com/istanbuljs/istanbuljs/issues/30#event-1044274942 is fixed

Stupidism:
It seems istanbul is the only good choice, for now.

test/mocha.opts
```diff
@@ -0,0 +1,3 @@
+--compilers js:babel-register
+--require ./test/setup
+src/**/*.test.js
```
JaKXz: should probably add the `--recursive flag` here

Stupidism: How? I didn't look at these file carefully because it's from your PR. @JaKXz

langpavel: I guess that it isn't needed because of `/**/`.. @JaKXz correct? Can you explain what is difference here?

package.json
```diff
@@ -198,8 +203,10 @@
     "lint:css": "stylelint \"src/**/*.{css,less,scss,sss}\"",
     "lint:staged": "lint-staged",
     "lint": "yarn run lint:js && yarn run lint:css",
-    "test": "mocha \"src/**/*.test.js\" --require babel-register --require test/setup.js",
+    "test": "mocha",
     "test:watch": "yarn run test -- --reporter min --watch",
+    "test:cover": "cross-env NODE_ENV=test nyc npm run test",
```
koistya: Unfortunately this will not work on Windows..

Stupidism:
Cau we finish this? @langpavel

langpavel:
@Stupidism Can you rebase please?
@frenzzy Can you review please?

.nycrc
```diff
@@ -0,0 +1,14 @@
...
+  "include": [
+    "src"
+  ],
+  "reporter": [
+    "lcov",
+    "text"
+  ],
+  "sourceMap": false,
+  "instrument": false
+}
```
frenzzy: Should we exclude `src/**/*.test.js` from coverage report or it is excluded by default?
For example: [src/components/Layout/Layout.test.js](https://github.com/kriasoft/react-starter-kit/blob/master/src/components/Layout/Layout.test.js)

Stupidism: image
I think it is excluded by default.

Stupidism:
I squashed all the commits @frenzzy @langpavel @koistya

test/mocha.opts
Stupidism: That's from RSK's old script https://github.com/kriasoft/react-starter-kit/pull/1213/commits/0025fa31f6ce333ac55a3298a59b1cb563560160#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L159

langpavel:
Nice, thanks!

PR #1230, opened 2017-04-20
chore(pre-commit): work with eslint --fix

Stupidism:


Stupidism:
```

  "//pre-commit": [
    "Special pre-commit working with `lint --fix` perfectly with a set of git commands",
    "[lint-staged] issue: Add better support for partially staged files",
    "https://github.com/okonet/lint-staged/issues/62#issuecomment-286830310",
    "[stackoverflow] question: how-can-you-remove-and-then-restore-non-index-changes-using-the-stash-without-ev",
    "http://stackoverflow.com/questions/13889242/26685296#26685296"
  ],
```

Stupidism:
Problem found:
[ ]: when there's nothing to stash and your stash list is not clear, `pos-stash` would pop a stash you don't want
[ ]: when you press `Ctrl+C` during commit, `pop-stash` would not pop for you, you can only manually do that

langpavel:
Why? Can you explain?

Stupidism:
I noticed `pre-commit: lint-staged` was removed in a commit without any explaination(also, dependency `pre-commit` was not removed, so maybe it's a mistake.): https://github.com/kriasoft/react-starter-kit/commit/8e28772eec4769fc8ab77b6f85d061e859e4148a.
Actually, I was kind of like this pre-commit hook but I encontered a problem while using it. 
If I want to commit partially, lint-staged would add my rest part into the commit. So I must `git stash --keep-index` first and than commit, but there will be conflicts everytime after the partial files are auto-fixed.

I looked for solution with google and found this: https://github.com/okonet/lint-staged/issues/62#issuecomment-286830310. 

I think it's a good thing to have a pre-commit hook that lints. We may have three options:
1. pre-commit: `lint-staged` (always breaks when there are lint errors even simple ones that can be auto-fixed)
2. pre-commit: `lint-staged --fix` (must `git stash --keep-index` before commit, and must handle conflicts after partial files are auto-fixed)
3. pre-commit: `stash-unstaged; lint-staged; pop-stash` (the thought is wonderful, but commands are not perfect so far. Maybe we should discuss more about this)

Stupidism:
Considering? @langpavel

frenzzy:
I think the solution is too complex to be used in multiple projects and also contains cross-platform issues like does not work on windows.

Maybe you can rewrite it in javascript and send PR to [lint-staged](https://github.com/okonet/lint-staged) instead?

langpavel:
Fixing this on `lint-staged` would be the best!

Stupidism:
Agreed, I'm suggesting to `lint-staged`'s author.

langpavel:
@Stupidism Can you paste issue/PR link here?

Stupidism:
https://github.com/okonet/lint-staged/pull/75
They already have this PR

Stupidism:
@langpavel new tool shown up: https://github.com/motiz88/git-exec-and-restage

langpavel:
@Stupidism Can you try it and open a new issue/PR?

PR #1237, opened 2017-04-24
Initialize a new HTTP client for each web request

koistya:
In the context of building isomorphic (universal) apps, it's better to initialize a new HTTP client for each web request during server-side rendering, it unlocks a couple of optimization techniques. This HTTP client can be used now as follows:

#### Before

```js
import fetch from '../../core/fetch';
export default {
  path: '/products',
  async action() {
    const resp = await fetch('/api/products');
    const data = await resp.json();
    return { title: 'Products', component: <Products {...data} /> };
  }
}
```

#### After

```js
export default {
  path: '/products',
  async action({ fetch }) {
    const resp = await fetch('/api/products');
    const data = await resp.json();
    return { title: 'Products', component: <Products {...data} /> };
  }
}
```

- [x] Add `src/createFetch.js` that provides deafult values for `fetch`
- [x] Remove `src/core/fetch` in favor to `isomorphic-fetch`
- [x] Initialize a new HTTP client utility (`fetch`) for each HTTP request, see `src/server.js`
- [x] Refactor `config.js` to have a single export, it's only supposed to be used on the server
- [x] Pass some config info from server to client, serialize it inside `Html` component into `window.App` variable
- [x] Remove `src/core` folder, it's better to put core framework modules (`router.js`, `history.js` etc.) right into the `/src` folder


langpavel:
Looks similar to one I have on redux branch, definitely better than global import. Have you HoC letting you access fetch in any component?

koistya:
Currently the `fetch` utility is passed via React's `context`:

```js
class MyComponent extends React.Component {
  static contextTypes = { fetch: PropTypes.func.isRequired };
  handleClick = (event) => {
    this.context.fetch('/api/products').then(x => x.json()).then(data => this.setState({ data });
  }
  render() { ... }
}
```

phpeek:
@koistya Do You plan to merge it ? This might be a BC break.

langpavel:
@phpeek Yes, this is breaking change, but I really like see this discussion again.
Singletons like fetch are evil especially on server.
Note that current implementation of fetch needs passing credentials manually if needed... now you can handle this on one place.

docs/data-fetching.md
```diff
@@ -4,10 +4,8 @@ There is isomorphic `core/fetch` module that can be used the same way in both
 client-side and server-side code as follows:
 
 ```jsx
-import fetch from '../core/fetch';
-
 export const path = '/products';
-export const action = async () => {
+export const action = async ({ fetch }) => {
```
langpavel: Example here is on router. But I believe users are importing fetch on more nested components so HoC implementation and React HoC component `withFetch()` which correctly bridge `context` will be needed for users.

src/client.js
```diff
@@ -29,6 +30,10 @@ const context = {
     const removeCss = styles.map(x => x._insertCss());
     return () => { removeCss.forEach(f => f()); };
   },
+  // Universal HTTP client
+  fetch: createFetch({
+    baseUrl: window.App.apiUrl,
+  }),
```
langpavel: add `credentials: 'same-origin'` here?

koistya:
@phpeek yes, we need to move the project forward, in all my universal web app projects I had to initialize a new HTTP client object for each web request on the Node.js side. This must be included into the boilerplate project in one way or another. And the remaining updates in this PR is just an attempt to simplify the project structure a little. It shouldn't be hard to merge them into your project, assuming you're using a good Diff and 3-way Merge Tool (GUI). What's your concern with that?

Shadowman4205:
@langpavel  So now how i will use fetch on redux actions?

langpavel:
@Shadowman4205 Same way as you did. `fetch` helper is still here.

Shadowman4205:
@langpavel  u mean i will use this way

> import fetch from '../createFetch';

Shadowman4205:
I tried this 

```
import createFetch from '../createFetch';
import config from '../config';

const fetch = createFetch({baseUrl: config.api.serverUrl});
```

and it say i cannot use config on client side.
so i tried

> const fetch = createFetch({baseUrl: null});
but i get 

> Error: only absolute urls are supported

langpavel:
@Shadowman4205 
You can use [`fetch` helper](https://github.com/kriasoft/react-starter-kit/blob/feature/react-intl/src/store/createHelpers.js#L35..L45) in Redux action like this:
```js
export function myAction(data) {
  return async (dispatch, getState, { client, history, fetch }) => { // see all helpers in src/store/createHelpers.js
    dispatch({ type: 'MY_ACTION_START' /* ... */ });
    // await fetch(...)
    dispatch({ type: 'MY_ACTION_SUCCESS' /* ... */ });
  }
}
```

Shadowman4205:
In my routes i doing dispatch(myAction)
but then i am getting

> TypeError: fetch is not a function

langpavel:
@Shadowman4205 Why you cannot write only `return async (dispatch, getState, { fetch }) => { ... }` ?

PR #1242, opened 2017-04-27
fix(server/jwt):handle 401 caused by secret change

Stupidism:
When config `auth.jwt.secret` is changed, this `UnauthorizedError: invalid signature` error happenens.
It's very serious because it can't be recovered automatically after refresh. User must clear cookie to continue to use the app.
```
UnauthorizedError: invalid signature
  
  - index.js:101 
    [react-starter-kit]/[express-jwt]/lib/index.js:101:22
  
  - index.js:54 
    [react-starter-kit]/[express-jwt]/[jsonwebtoken]/index.js:54:18
  
  - next_tick.js:67 _combinedTickCallback
    internal/process/next_tick.js:67:7
  
  - next_tick.js:98 process._tickCallback
    internal/process/next_tick.js:98:9
```

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -54,6 +54,16 @@ app.use(expressJwt({
   credentialsRequired: false,
   getToken: req => req.cookies.id_token,
 }));
+// Error handler for express-jwt
+app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
+  if (err instanceof Jwt401Error) {
+    console.error('[express-jwt-error]', req.cookies.id_token);
+    // `clearCookie`, otherwise user can't use web-app until cookie expires
+    res.clearCookie('id_token');
+  }
+  next(err);
+});
+
 app.use(passport.initialize());
 
 if (__DEV__) {
@@ -146,17 +156,6 @@ app.get('*', async (req, res, next) => {
 const pe = new PrettyError();
 pe.skipNodeFiles();
 pe.skipPackage('express');
-
-// express-jwt error handler
-app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
-  if (err instanceof Jwt401Error) {
-    console.error('[express-jwt]', req.cookies.id_token, pe.render(err));
-    res.clearCookie('id_token');
-    res.redirect('/');
-  } else {
-    next(err);
-  }
-});
 
 app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
   console.error(pe.render(err));
--- a/src/server.js
+++ b/src/server.js
@@ -60,11 +60,6 @@ app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
     console.error('[express-jwt-error]', req.cookies.id_token);
     // `clearCookie`, otherwise user can't use web-app until cookie expires
     res.clearCookie('id_token');
-    const jwtError = new Error(err.message);
-    jwtError.status = err.status;
-    jwtError.name = err.name;
-    jwtError.stack = err.stack;
-    next(jwtError);
   } else {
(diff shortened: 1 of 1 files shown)


Stupidism:
### How to reproduce
1. Open chrome - F12/inspect - Application - cookies
2. Input a cookie key-value with key `id_token` and value `(some valid jwt)`

Notice: If you input an invalid jwt, you will get another error and it can also be handled by this

src/server.js
```diff
@@ -147,6 +147,17 @@ const pe = new PrettyError();
...
+// express-jwt error handler
+app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
+  if (err instanceof Jwt401Error) {
+    console.error('[express-jwt]', req.cookies.id_token, pe.render(err));
+    res.clearCookie('id_token');
+    res.redirect('/');
+  } else {
+    next(err);
+  }
+});
```
frenzzy: Maybe move it right after `Authentication` middlewares and render a page for unauthorized user instead of redirecting to the main page?

Stupidism: This error only happens in two situations:
1. Dev Operator changed `auth.jwt.secret` on the server side
2. User modified cookie `id_token` on the client side

I thought neither of these is a normal process for user to use a webapp, so I just clear the cookie and redirect to homepage(let user login).

As I can see now, only `clearCookie` is necessary, and then `next(error)` let the final errorHandler to handle it?

Stupidism:
This error only happens in two situations:
1. Dev Operator changed `auth.jwt.secret` on the server side
2. User modified cookie `id_token` on the client side

I thought neither of these is a normal process for user to use a webapp, so I just clear the cookie and redirect to homepage(let user login).

As I can see now, only `clearCookie` is necessary, and then `next(error)` let the final errorHandler to handle it?

src/server.js
```diff
@@ -54,6 +54,16 @@ app.use(expressJwt({
...
+// Error handler for express-jwt
+app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
+  if (err instanceof Jwt401Error) {
+    console.error('[express-jwt-error]', req.cookies.id_token);
+    // `clearCookie`, otherwise user can't use web-app until cookie expires
+    res.clearCookie('id_token');
+  }
+  next(err);
+});
+
```
koistya: @Stupidism what do you think about moving this code into the existing error handler at the bottom of this file?

Stupidism: I think it's more associated with express-jwt. The problem is not to display error message, but recover(clearCookie) after error happens. 

https://github.com/auth0/express-jwt#error-handling

frenzzy: And in case of jwt error we should just continue rendering a page the same was as for unauthorized user, so after `clearCookie` just `next()` without arguments and `return`.

Stupidism: So the error is swallowed, I'd rather present it in errorPage.
If someone want it to be swallowed like the code I wrote in first commit, he can change it easily.
But if we swallow it directly, he might won't know there's an error at all.

frenzzy: Maybe set 401 status then?

frenzzy: @koistya all auth middlewares must go one by one and before any others including this one with auth errors handling, don't you think?

Stupidism: As I tested, it is 401 because of this [code](https://github.com/Stupidism/react-starter-kit/blob/e07468646c5d7c3d452ea320eae4d6fd07be55e6/src/server.js#L171).
If you want it more explicit, I can do it

koistya: @Stupidism what do you think about adding `res.status = 401` right after `res.clearCookie('id_token')`?

Issue #1241, opened 2017-04-26
Performance issues with vendor bundles


ghost:
Hi there! I really love this repo. Really, thanks for all the work.

I started off building an application using this boilerplate (its only static content that's live right now), but [performance has suffered.](https://developers.google.com/speed/pagespeed/insights/?url=socialworkschi.org&tab=mobile)

Given the boilerplate code, I'd love to get input in optimizing compression and reducing build sizes. I'm poking around but would love some feedback.

Andriy-Kulak:
+1. I have a similar bad performance numbers on my site and I only added a few packages and have a dozen or two components. I noticed that th**e React Starter Kit website** **(Screenshot Below)** has considerably better numbers (70/100 according to google), using the same boilerplate I assume. Anyone have any ideas?

![screen shot 2017-04-26 at 9 00 54 pm](https://cloud.githubusercontent.com/assets/12179976/25463377/c4711e8e-2ac3-11e7-879b-ba40da335381.png)

buildbreakdo:
@Fahd You have some huge custom font files included in your clientjs that look like they are in base64 format, making them even bigger.

Anywho, compression can be handled by file hosting services like S3 (make sure to turn gzip on in the bucket configuration on S3) for assets like `client.js`, `vendor.js`, custom font files, etc as part of your deploy script you would do an `aws s3` upload of the files once a production build has run, when the app runs in production mode you would use your s3 bucket address, dev mode would just use the local public directory.

ExpressJS also has a compression library that is really easy to use. Used to be included in the base project but they pulled it out.

Found here https://github.com/expressjs/compression and works like:
`npm install compression --save`

```
// server.js
...
var express     = require('express')
var compression = require('compression')

var app = express()
app.use(compression())
...
```

You should really only be using this if you do not have a proxy server like Apache or NGINX in front of your node app. Otherwise, you should turn on gzip compression from Apache or NGINX respectively.

@Andriy-Kulak Caching is set in the response header, so in server.js you would add:
```
// Server.js
...
res.set('cache-control', 'private, max-age=86400');
...
```
Which would cache the response for one day (like what the Google homepage does). See pull request https://github.com/kriasoft/react-starter-kit/pull/1245 .

langpavel:
This starter kit is best in isomorphic. Composition should be handled on proxy especially if you scale horizontally. But there are more recently things you should reach recently and you must deal with.. like any other starter. So it always not be the best dealing with this from beginning

[...]

PR #1245, opened 2017-04-27
Default response header Cache-Control to 1 day in production;disable response cache in __DEV__

buildbreakdo:
Default Cache-Control to 1 day in production (adjust per needs, 1 day is just a generic default pulled from Google's home page). When in `__DEV__` mode disable caching of responses. Issues are touched on in #1241 .


koistya:
What if data API changed, and the prev. version of the JavaScript bundle can't handle it. I assume this technique must be combined with server version checking (via Ajax).

buildbreakdo:
@koistya Intriguing question, according to this: http://stackoverflow.com/questions/23603023/file-caching-query-string-vs-last-modified (top answer last scenario) the RSK is already in best practice space. The url to the file changes. So there is nothing to manage on the RSK end, it is already managed by the bundling process through generating a new file name and thus a new url for that file.

All this does, is cache that response if it is asked for again. If the file URL changes then that will be a new response, and not pulled from the cache.

buildbreakdo:
@koistya Confused by all the drinks I had, drinking and coding is not always the best of ideas, but it is fun!

Okay, what should be happening in a scenario where the API version changes and the prev version of the JavaScript bundle cannot handle it is API versioning. The API version would be hard coded into the client.js bundle via the URL, like `reddit.com/api/v1/friends` (https://www.reddit.com/dev/api/). If the a new version of the API is published, the old bundle that relies on a particular API version to work would still be targeting the old API version at `/api/v1` while the new version of the API would be published at `api/v2`. Personal preference is for URL based versioning, for the sake of simplicity and readability, plenty of religious discussions around that.

Alternatively, could publish a new client and API simultaneously to keep them in sync, which is what we traditionally do in most web app scenarios where the API is included in the same project (assuming you're talking about situations where the API has been externalized like https://github.com/kriasoft/nodejs-api-starter).

ulani:
@buildbreakdo  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1246, opened 2017-04-27
Facebook: request displayName profile field

jorrit:
`profile.displayName` will be undefined when the displayName field is not explicitly requested.


jorrit:
Thanks!

PR #1250, opened 2017-04-30
Integrate Relay "Modern"

koistya:
### Getting Started with Relay "Modern"

1. Copy and paste your GraphQL schema to `src/schema.graphql` (see [example](https://reactstarter.com/graphql/schema))

2. Decorate your React components with either `createFragmentContainer`, `createRefetchContainer` or `createPaginationContainer` wrappers from `react-relay` module.

3. Write a root query inside your route handler(s), import necessary fragments, and call `api.fetchQuery(query)`.

4. Run `yarn run relay` (or `yarn run relay -- --watch`) that extracts GraphQL fragments from your code and creates `__generated__/*.graphql.js` files

5. Run `yarn start` to start your project as usual.

### Relay Component Example

```js
import React from 'react';
import { graphql, createFragmentContainer } from 'react-relay';

class NewsFeed extends React.Component {
  render() {
    return (
      <div>
        {this.props.news.map(x => (
          <section>
            <h2>{x.title}</h2>
            <p>{x.summary}</p>
          </section>
        ))}
      </div>
    );
  }
}

export default createFragmentContainer(News, graphql`
  fragment NewsFeed_news on NewsItem @relay(plural: true) {
    title
    summary
  }
`);
```

### Route Example

```js
import React from 'react';
import { graphql } from 'relay-runtime';
import NewsFeed from './NewsFeed';

export default {
  path: '/news',
  async action({ api }) {
    const data = await api.fetchQuery(graphql`
      query newsQuery {
        news {
          ...NewsFeed_news
        }
      }
    `);
    return { title: 'News Feed', component: <NewsFeed news={data.news /> };
  }
}
```

Note, you can use the same `api` utility to send normal Ajax requests, for example `const news = await api.fetch('/api/news').then(x => x.json())`

### More Info

* [Getting started with Relay "Modern" for building isomorphic web apps](https://medium.com/@tarkus/getting-started-with-relay-modern-for-building-isomorphic-web-apps-ae049e4e23c1)
* https://facebook.github.io/relay/docs/relay-modern.html

**P.S.** Those of you who are not using either Relay or GraphQL or both, I will add a "feature" PR that removes Relay from the project, so you could easily remove Relay integration in case you don't ready to use it yet.


stubailo:
Hey have you considered using Apollo as well? I think it could be a better fit for the project.

stubailo:
Also, it looks like with this syntax where you pass in the data with a promise from fetchQuery, you won't receive updates from the store.

stubailo:
(Plus, Apollo includes server-side rendering support out of the box, so it should work with universal router with no special modifications, and people don't need to call a special API in the router)

jonirrings:
@stubailo seems there is a branch named `feature/apollo` ,but the server-side rendering not included

langpavel:
@jonirrings @stubailo SSR is supported and is very efficient, because it maps query directly to graphql query without need of HTTP roundtrip

stubailo:
Aside from the point of Relay vs. Apollo, I'd like to reinforce the point that with this approach, people won't be able to properly incorporate mutation results into their components.

langpavel:
> Aside from the point of Relay vs. Apollo, I'd like to reinforce the point that with this approach, people won't be able to properly incorporate mutation results into their components.

@stubailo Can you explain this a little?

Note: I'm very happy with Apollo and everything it offers. I don't really see any new benefits of Relay Modern against Apollo. May be I'm only less educated in Relay..

koistya:
@stubailo thanks for the feedback, much appreciated! I'm not familiar with Apollo client good enough, but it seems like it misses some core features like being able to extract queries from the code and use query IDs for `client<->server` communication, resulting in smaller bundles and more efficient operation (CPU/memory), especially on mobile devices. Also, Relay Modern allows using its store for client-only states, which may replace the need in Flux/Redux in some cases.

langpavel:
@koistya did you think about feature/relay branch?
Make this PR tracking like I did with redux and mention this in readme..
We can gather some feedback after which approach is better for various situations

stubailo:
>  I'm not familiar with Apollo client good enough, but it seems like it misses some essential features (in comparison to Relay) like being able to extract queries from the code and use query IDs for client<->server communication, resulting in smaller bundles and more efficient operation (CPU/memory), especially on mobile devices. Also, Relay Modern allows using its store for client-only states, which may replace the need in Flux/Redux in some cases.

All of these are present in Apollo as well, in fact it supported those features much earlier! Also, there are a lot of things Apollo does that Relay doesn't. But anyway for one of the features you can see this post: https://dev-blog.apollodata.com/persisted-graphql-queries-with-apollo-client-119fd7e6bba5

> people won't be able to properly incorporate mutation results into their components.

It looks like you're using a fetching function to return data with a promise. And the promise result fires only once, meaning it won't fire again when the store updates after a mutation.

stubailo:
Happy to get on a call anytime to disambiguate, part of the reason I wanted to bring it up is that there are a lot of misconceptions about Relay vs. Apollo features and benefits, so I was worried you hadn't taken a close look at the possible benefits of using Apollo.

jonirrings:
@koistya query with `query ID` is supported in Apollo months earlier before `relay modern` release , @stubailo is it aliased as `static query`?
Personally, I use apollo and relay both, when work with redux I prefer apollo, without redux, I choose relay.
apollo is more active on experimental spec support of GraphQL and more documented, as for relay, the classic version is too magic, and I'm still study the modern one, which is more understandable, thanks for @stubailo your `relay` work, I've take much code from your work:stuck_out_tongue_winking_eye:

stubailo:
People usually call that feature persisted queries.

Falieson:
> Personally, I use apollo and relay both, when work with redux I prefer apollo, without redux, I choose relay.
@jonirrings

jonirrings:
@Falieson expand on which? 
for working with `apollo`, it is easy to integrate with redux.
for working with `relay`, I use the `isomorphic-relay` solution, without redux.
if working with `redux` and `relay classic`, I guess this may tackle with your problem
···js
<Provider store={store}>
     <RelayRootCantainer(XXX)/>
</Provider>
...
export default connect(mapStateToProps,mapDispatchToProps)(Relay.CreateContainer(YYY))
``` 
for `relay modern` I'm struggling with it.

Falieson:
all: has anyone figured out the fix to make the build pass for this PR?

@jonirrings (this discussion is probably better elsewhere but oh well) I'm curious what you use redux for? My primary motivation for redux is to be able to create independent UI modules that can optionally talk to each other. for example I am making "px-shell" which does layout/nav/theme/notifications/toasts and any other "px" ui modules knows how to get theme info from the px-shell and dispatch to notifications/toasts

langpavel:
@Falieson redux can be discussed at #1084

zomco:
I want to use `createPaginationContainer` instead of `createFragmentContainer` in example, but it occurs error `Cannot read property 'xxx' of undefined`.
properties defined in `Query` seems not to be passed to `Fragment` , any workaround?

cruisecoders:
How we can add mutations in this

langpavel:
@cruisecoders Mutation is not different from query on server. The fact that it is mutation — this is only relevant for client. By convention.
Query can yield with different data, from mutation it is expected.
Clients are written with this, but they can refetch in both cases...
It's only on you when you decide.. Mutation is obvious, query is stable — for most situations — if it is `volatile` then you need `subscribe` mechanism

All this is essentials, you should read more about.

BTW mutations are described well at the source of idea. This kit is very good at pointing you the right way. If you don't want magick you should try feature branches. If you don't feel lucky with us, you can teach with create-react-app which now looking for Relay. But I must warn you, it cannot be without you-must-try-yourself moments like with this kit. We are maintaining RSK for best productivity with newest packages, so if you want new app, you should fork and continuously merge our branches, because ejecting create-react-app is not long-living solution. You can fork feature branch and we will maintain the core, you can change but you should merge and this is the path..

Relay is technology which is too magic for me. I don't know if @koistya have more experience, but if you want trivials — like mutations are — there is no best way at the moment..

You can choose Apollo with web sockets for transport to subscriptions or you can be fulfilled with pushing this into Redux via [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) channel for example..

DaveyEdwards:
@cruisecoders I have a working example of Koistya's Relay Modern branch with a create mutation (its outdated compared to where react-starter-kit currently is).  Hoping to soon do a update of my project with either adding Relay Modern to RSK, or go the nodejs-api-starter + separate relay frontend.

As @langpavel was implying its not hard to implement them (when you find a working example to reference that is)

Here is my project: https://github.com/DaveyEdwards/myiworlds

Here is my working create mutation: https://github.com/DaveyEdwards/myiworlds/tree/master/src/routes/create
To show you it working (looks basic because it was just to get minimal working product)


Inside my queries folder is the database queries that will do all the mutations to the database.  They are all working with the database and GraphQL, I just need to create the Relay Modern mutations and hook them up
https://github.com/DaveyEdwards/myiworlds/tree/master/src/data/GoogleCloudPlatform/StorageAndDatabases/Datastore/Circle/Queries

I am going to be making youtube videos on all aspects of my project (uses React/RelayModern/GraphQL/Google Datastore) to help others, if there is any specific areas you are confused with on Relay Modern shoot me a message.  I will try and work it into my tutorials

zomco:
How to do unit test in this

ulani:
@koistya  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1269, opened 2017-05-15
chore(pre-commit): use git-exec-and-restage

Stupidism:


langpavel:
Hi @Stupidism 

Because this is a sensitive thing, do you consider this safer than what we have now?
Can you sum (or copy) benefits here? — best is in your opening message. This is important part of contributing workflow too :wink: 
Work this across all platforms (win, lin, osx)?

Thank you!

Stupidism:
@langpavel As I tried last time, this repo had some small problems. I listed them for in the issue https://github.com/motiz88/git-exec-and-restage/issues/5

1. blank commit problem(This barely happens actually) https://github.com/motiz88/git-exec-and-restage/issues/7
2. could not work fine with partially staged file(better than what we had before, because they only add fully staged files after `lint --fix`) and they labeled it as `wontfix`(Because to achieve this, unsafe `git stash` must be involved) https://github.com/motiz88/git-exec-and-restage/issues/8
3. Although fullyAdded files are git-added after fix, commit content in preview(when you input commit message) is still not fixed:

About across platforms, I don't have experience about this and I don't have so many computers.

Stupidism:
Have you tried to commit sth by yourself? @langpavel

langpavel:
@Stupidism Not yet

Issue #1274, opened 2017-05-21
Flexboxgrid media rules doesn't import globally from react-starter-kit


Blitss:
I'm trying to import `flexboxgrid` css module from starter-kit. 
I've added import rule to Layout.css:
`@global-import 'flexboxgrid' ;`
But media rules are using local names:
![f4f7617e18e54b34879e7d44eefdf3b9](https://cloud.githubusercontent.com/assets/17164985/26282314/f6b3df8c-3e17-11e7-8b5c-32a3711721fb.png)
Here's an example: https://github.com/Blitss/react-starter-kit
What can I do to fix it? I'm using node v7.9.0 on Windows

frenzzy:
You can specify two css loaders in your `tools/webpack.config.js` for local and global styles separately like so:
```js
{
  test: /\.css$/,
  include: [path.resolve(__dirname, '../src')], // Internal Styles
  use: [
    { loader: 'isomorphic-style-loader' },
    { loader: 'css-loader', // CSS Modules Enabled
      options: {
        importLoaders: 1,
        sourceMap: isDebug,
        modules: true,
        localIdentName: isDebug ? '[name]-[local]-[hash:base64:5]' : '[hash:base64:5]',
        minimize: !isDebug,
        discardComments: { removeAll: true },
    } },
    { loader: 'postcss-loader', options: { config: './tools/postcss.config.js' } },
  ],
},
{
  test: /\.css$/,
  exclude: [path.resolve(__dirname, '../src')], // External Styles
  use: [
    { loader: 'isomorphic-style-loader' },
    { loader: 'css-loader', // CSS Modules Disabled
      options: {
        sourceMap: isDebug,
        minimize: !isDebug,
        discardComments: { removeAll: true },
      } },
  ],
},
```
Then import external css inside your component:
```js
import React from 'react';
import withStyles from 'isomorphic-style-loader/lib/withStyles';
import external from 'external.css';
import s from './MyComponent.css';

class MyComponent extends React.Component {
  render() {
    return <div className={s.local}><div className="global">Hi</div></div>
  }
}

export default withStyles(external, s)(MyComponent);
```

langpavel:
Merge latest changes and import your file from JS, like normalize.css is now. Do not @import from css!

PR #1277, opened 2017-05-23
Update packages

langpavel:
- [x] `yarn upgrade`
- [x] manually upgrade packages with major version change

## Removed
I must remove `postcss-global-import`. See scherebedov/postcss-global-import#2

## Upgraded packages
+    "normalize.css": "^7.0.0",
+    "autoprefixer": "^7.1.1",
+    "cross-env": "^5.0.0",
+    "eslint-config-airbnb": "^15.0.1",
+    "eslint-plugin-jsx-a11y": "^5.0.3",
+    "eslint-plugin-react": "^7.0.1",
+    "pleeease-filters": "^4.0.0",
+    "postcss": "^6.0.1",
+    "postcss-calc": "^6.0.0",
+    "postcss-color-function": "^4.0.0",
+    "postcss-custom-media": "^6.0.0",
+    "postcss-custom-properties": "^6.0.1",
+    "postcss-custom-selectors": "^4.0.1",
+    "postcss-flexbugs-fixes": "^3.0.0",
+    "postcss-import": "^10.0.0",
+    "postcss-loader": "^2.0.5",
+    "postcss-media-minmax": "^3.0.0",
+    "postcss-nested": "^2.0.2",
+    "postcss-nesting": "^4.0.1",
+    "postcss-selector-matches": "^3.0.1",
+    "postcss-selector-not": "^3.0.1",


tools/webpack.config.js
```diff
@@ -275,7 +277,7 @@ const serverConfig = {
         ...rule.query,
         presets: rule.query.presets.map(preset => (preset[0] !== 'env' ? preset : ['env', {
           targets: {
-            node: parseFloat(pkg.engines.node.replace(/^\D+/g, '')),
+            node: pkg.engines.node.match(/(\d+\.?)+/)[0],
```
frenzzy: why don't you like replace? just remove `parseFloat`

tools/postcss.config.js
```diff
@@ -16,7 +16,9 @@ module.exports = () => ({
     // Transfer @global-import rule by inlining content with :global CSS Modules scope
     // e.g. @global-import 'draft-js/dist/Draft.css'
     // https://github.com/scherebedov/postcss-global-import
-    require('postcss-global-import')(),
+    // BUGGY: See https://github.com/scherebedov/postcss-global-import/issues/2
+    // require('postcss-global-import')(),
+    //
```
frenzzy: You need also remove `@global-import` usages from css.
Maybe completely remove the mention of this plugin in favor of [two loaders](https://github.com/kriasoft/react-starter-kit/issues/1274#issuecomment-302958948)?

tools/webpack.config.js
langpavel: `babel-preset-env` [have new check](https://github.com/babel/babel-preset-env/blob/5ff37e315115910cf96f6a766cf3531912796a81/src/targets-parser.js#L49..L64)

tools/postcss.config.js
langpavel: [Two loaders](https://github.com/kriasoft/react-starter-kit/issues/1274#issuecomment-302958948) looks better :+1:

tools/webpack.config.js
langpavel: This will extract first version string if you write more versions

frenzzy: what am I missed? https://jsfiddle.net/85jaewyv/

langpavel: https://jsfiddle.net/langpavel/bo5e1c33/

langpavel:
@frenzzy It looks that the [last commit](https://github.com/kriasoft/react-starter-kit/pull/1277/commits/391948dae3c5d5eeb3277e0349a12da0d14e3d19) does not work. Can you look what is bad with them?

tools/webpack.config.js
```diff
@@ -77,7 +77,11 @@ const config = {
         },
       },
       {
+        // Internal Styles
         test: /\.css/,
```
frenzzy: `/\.css$/`

src/components/Layout/Layout.css
```diff
@@ -8,8 +8,7 @@
  */
 
 @import '../variables.css';
-
-@global-import 'normalize.css' ;
+@import 'normalize.css' ;
```
frenzzy: Maybe remove it from `Layout.css` and add to `Layout.js` as an example?

tools/webpack.config.js
langpavel: Grrr

src/components/Layout/Layout.css
langpavel: Is it better?

langpavel:
@frenzzy last commit still not work.

frenzzy:
What exactly does not work? App is running and styles applied for me in your branch

langpavel:
@frenzzy I still see `.Layout-customClass-3Crft` instead of my `.customClass` inserted into `node_modules/normalize.css`

src/components/Layout/Layout.css
frenzzy: It doesn't matter for normalize because there are no class names inside. But it could be an example how users can add any other third-party css libs.

frenzzy:
Oh, of cause if you want global scope in external css you need to import them directly in js instead of `@import` (which is just concatenate files and final result is processed with local css modules scope)

langpavel:
That makes sense.. Hmm, so I will try

langpavel:
Ok @frenzzy, thank you for help! Now it works as expected. Can you review my last commit?
Should I wait for @koistya when he approve?

PR #1279, opened 2017-05-25
Router receive whole react context

langpavel:
This looks like trivial enhancement but when you extending context, you will see repeating yourself all the stuff from context to router..


frenzzy:
Or alternatively, the context can be passed to the router's constructor once:
```js
const router = new Router(routes, { context });

// ...
router.resolve({ path, query });
```

langpavel:
Oh @frenzzy I was in hurry :-)
Is there real difference?

frenzzy:
only aesthetic difference :)

PR #1301, opened 2017-06-09
Include correct userAgent in server for react server-side rendering

inspiredjw:
This helps to have correct auto-prefixed server rendered page.


inspiredjw:
This fix resolved these problems.

https://github.com/callemall/material-ui/issues/4437
https://github.com/callemall/material-ui/issues/6920
https://github.com/callemall/material-ui/issues/2336

langpavel:
That way `global.navigator.userAgent` will not be stable on entire request lifecycle.
Because every new request will change global state, you can observe glitches if you make multiple parallel requests across different browsers.
This is really bad idea.

PR #1317, opened 2017-06-18
Isomorphic Hot Module Replacement

frenzzy:
- add hot module replacement support for server-side rendering
- use single developer server with Browsersync and without proxy, default port is `3000`
- add `--silent` flag to avoid default browser opening (`yarn start -- --silent`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ and newcomers to the industry.
 
 <p align="center" align="top">
   <a href="https://rollbar.com/?utm_source=reactstartkit(github)&amp;utm_medium=link&amp;utm_campaign=reactstartkit(github)"><img src="https://koistya.github.io/files/rollbar-362x72.png" height="36" align="top" /></a>
-  <a href="https://x-team.com/hire-react-developers/?utm_source=reactstarterkit&utm_medium=github-link&utm_campaign=reactstarterkit-june"><img src="https://koistya.github.io/files/xteam-255x72.png" height="36" align="top" /></a>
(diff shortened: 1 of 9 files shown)


WiktorKa:
After merging this commit, I had to make some changes to `src/core/createApolloClient/createApolloClient.client.js` - I had to add `http://localhost:3000` to `uri` param - otherwise I was getting `only absolute urls are supported` error from component that was fetching data via Apollo. I guess my solution is not production ready, right?

```
networkInterface: createNetworkInterface({
    uri: 'http://localhost:3000/graphql',
  }),
```

frenzzy:
@WiktorKa [server-side createFetch()](https://github.com/kriasoft/react-starter-kit/blob/42ec07210eb384ab48120ac4b24d1ae1d4edd420/src/server.js#L113) must use [`src/config.js#api.serverUrl`](https://github.com/kriasoft/react-starter-kit/blob/42ec07210eb384ab48120ac4b24d1ae1d4edd420/src/config.js#L24-L25) as a default `baseUrl` with ability to modify `API_SERVER_URL` environment variable for production.

WiktorKa:
@frenzzy It is configured like you said. I reverted my change and now I get `Error: Invariant Violation: Argument of /assets/src/routes/home/home.graphql?bc804126 passed to parser was not a valid GraphQL DocumentNode. You may need to use 'graphql-tag' or another method to convert your operation into a document` so I'm guessing it may have smth to do with the new webpack configuration... ;/

frenzzy:
You need a [loader for such files](https://github.com/kriasoft/react-starter-kit/blob/ba4894a069638d1675c0f6683356fd33c39de073/tools/webpack.config.js#L158-L162) and also need to add the same extension to [exclude list of file-loader](https://github.com/kriasoft/react-starter-kit/blob/42ec07210eb384ab48120ac4b24d1ae1d4edd420/tools/webpack.config.js#L208-L217).

PR #1346, opened 2017-07-02
Intl rework

langpavel:
I want make localization more friendly for developers *and* for translators too.
What is missing?
What is required?
# Near future plan:
- All locale data will still be in one file, this is better for external translators and deployment.
- Default messages will be in hierarchy following `babel-plugin-react-intl`. This will not have an impact except if you add `build/messages` to versioning. **I need more opinion from you on this!**
- I want start new tool for translators on **port 3003 in development**. This tool should be very friendly, helpful with localization.
It should be very user friendly, showing primarily untranslated messages from current page, pointing out similarities and related messages from other pages. Realtime preview is the must, etc..


langpavel:
## Ports in development:
- [x] 3000 app itself
- [x] 3001 raw app
- [x] 3002 browsersync
- [ ] 3003 — **translation tool :tada: — write your ideas in this PR**

PR #1425, opened 2017-10-06
[feature/apollo] Single source of graphql query

KrzysztofKarol:
If we are using same query for SSR and on the client site (eg. in refetchQueries) I think maintaining single query is easier and DRYer.

### DIFF
--- a/src/components/Layout/Layout.test.js
+++ b/src/components/Layout/Layout.test.js
@@ -13,6 +13,7 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
 import configureStore from 'redux-mock-store';
+import { ApolloClient } from 'apollo-client';
 import thunk from 'redux-thunk';
 import App from '../App';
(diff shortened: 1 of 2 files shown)


langpavel:
All thumbs up :+1: but can you describe this more?

KrzysztofKarol:
Currently news are fetched twice. Once by Server and once by Client.

I don't understand why. But if we are using it twice it should has same source.

Let's assume we are reusing this query
1. Getting all news (SSR)
2. Updating list of news (Client)

```
const addNews = gql`
  mutation addNews...
`;

@graphql(addNews, { name: 'addNews' })
...
this.props.addNews({
 variables: {
    ..
  },
  refetchQueries: [{ query: newsQuery }],
})
```

langpavel:
Hi. I altered code a little.

Now router uses Apollo directly and serialized apollo cache is restored on client, so much better I hope :slightly_smiling_face:

Issue #499, opened 2016-03-12
Where do I load a CSS framework?


alexdrans:
I am attempting to load Semantic UI into the starter pack, and at the moment I am including it on all pages in the old fashion way. Where should I be loading my .css and .js files?

koistya:
@alexdrans try to configure a loader for these files, then import them from your custom CSS code, for example:

```css
@import '../../../node_modules/module-name/file.css';
```

Alternatively, you can specify the list of loaders along with their settings on the same line where you reference a CSS file, for example:

```css
@import '!!isomorphic-style-loader!css-loader!../../../node_modules/module-name/file.css';
```

ulani:
@alexdrans   thank you very much for crating this issue! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

NOTE: The `main` branch has been updated with React Starter Kit v2, using JAM-style architecture.

PR #1455, opened 2017-11-04
Feature/semanticui

ahmdilloo:
Hi
I added semantic ui to this project and use it in header and login page. i think maybe some one need this and can quickly use it.
like this: [#499](https://github.com/kriasoft/react-starter-kit/issues/499)


tim-soft:
Why not install semantic-ui to `node_modules`?

zocky:
@tim-soft - That only installs the js. Css still needs to be installed with the semantic-ui-css package. The semantic-ui-react instructions then say that the css should be imported using `import 'semantic-ui-css/semantic.min.css';` in index.js (which I suppose is client.js). This doesn't have the intended effect though.

How should that be done?

kalin-mv:
Hi,
please try this:

`yarn add semantic-ui-css`

**Layout.js**
```
import g from '../../../node_modules/semantic-ui-css/semantic.min.css';
import s from './Layout.css';

class Layout extends React.Component {
}

export default withStyles(g, s)(Layout);
```

tim-soft:
@kalin-mv this is a bad idea as it will inline a very, very large amount of CSS into your JS which gets sent to the client in one large chunk.

The better approach imo would be to use some webpack loader for the styles in combination with extract-text-plugin, mostly because of how absolutely large semantic-ui resources can be. Doing it this way would allow semantic css to get downloaded _in parallel_ to JS as it's own file, get cached by the browser and not bottleneck JS loading/execution.

langpavel:
As a PR this is really bad idea.. and outdated.
Everybody reaching this, please follow @kalin-mv's example

langpavel:
@kranz It depends.. If semantic-ui-css uses jquery only in `componentDidMount` and didn't try do anything on server, it should be OK

kranz:
@langpavel please apologize. I should have posted this question in semantic-ui-react...

PR #1466, opened 2017-11-23
Simplify server side webpack config

zation:
we can use `emitFile: false` to make server side not to generate static files

### DIFF
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -433,8 +433,7 @@ const serverConfig = {
           ...rule,
           options: {
             ...rule.options,
-            name: `public/assets/${rule.options.name}`,
-            publicPath: url => url.replace(/^public/, ''),
+            emitFile: false,
           },
         };
       }


langpavel:
Can you provide more info please? Whats going on when you apply this?

zation:
Server side should use the static files generated by client webpack, not server webpack. So we should make the server webpack `emitFile: false`. So that server webpack won't generate the duplicated and useless static files.

zation:
Previously, the config make server side and client side having the same path. So that there won't be duplicated files appears. But we can just use `emitFile: false` for server webpack to make this happen which also would be simple.

langpavel:
Thanks @zation,
can you tell me please, how diverged (`module.server.js`/`module.client.js`) files will work in this setup? Will this be ok?

zation:
This config wont effect js files. This will only effect static files such as images, svgs... If your diverged (`module.server.js`/`module.client.js`) files means js output, they have no relation ship with this config. They will only relate to the `output` in the root node of webpack config

frenzzy:
Need to check here if server continues return correct paths to media files during SSR after this change.
For example `` must render something like `` on client and absolutely the same `src` during server side rendering.

zation:
Yes, you can try the new config in all examples and they are absolutely the same between client and server. Since client and server have absolutely the same `publicPath` config in webpack. Let me explain more about why previous solution is duplicated:
The client and server `publicPath` is `/assets/` (Please check https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L45), and both client and server haven't rewrote it.
In the previous server webpack config, we config the pathname as `public/assets/${rule.options.name}`, and then remove "public"... Is this a duplicated config? https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L436
And of course, all the static files should generate from one side (client). So the server side config should `emitFile: false`.

frenzzy:
Thank you!

PR #1495, opened 2018-01-05
Abort build when there are compilation errors

jorrit:
I accidentally pushed a build to production that contained webpack compilation errors. As can be read at https://webpack.js.org/api/node/#error-handling, the `err` parameter covers just a portion of the possible errors that are possible during webpack compilation.


jorrit:
Thanks!

PR #1496, opened 2018-01-05
[feature/Apollo] Log Apollo Ops in Development

tim-soft:
Restores logging capability on all Apollo GraphQL operations while in development mode. 

The logging output is in the style of redux-logger, so the same as with redux ops, but also includes the time taken by the request.

capture


langpavel:
Thank you again :slightly_smiling_face:

PR #1502, opened 2018-01-13
Short circuit fetch for `/graphql` on SSR

nbclark:
The starter template with graphql uses node-fetch to route HTTP requests when doing SSR. For graphql queries, this leads to a decent bit of overhead as it issues HTTP requests when the request could be handled in the same process without the loopback http call. In our application, this leads to 50-100ms performance improvements per graphql query.

Since we are internally routing, mapping of request properties will need to be handled as needed. For example, we add a user to the request object based on the JWT, so that needs to be passed through. We accomplish that through additional args to the createFetch method.

Also, there is a very lightweight shim for the response object (json()) to support backwards compat.

This could be extended to do something similar for API calls.

Happy to make any requested changes.

Thanks for a great starter template.

### DIFF
--- a/src/createFetch.js
+++ b/src/createFetch.js
@@ -36,7 +36,7 @@ function createFetch(fetch: Fetch, { baseUrl, cookie, schema }: Options) {
     },
   };
 
-  return (url: string, options: any) => {
+  return async (url: string, options: any) => {
     const isGraphQL = url.startsWith('/graphql');
     if (schema && isGraphQL) {
       // We're SSR, so route the graphql internall to avoid latency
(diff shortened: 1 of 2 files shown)


langpavel:
Hi. I only add note here, that on `feature/apollo` additional HTTP request is avoided by using `SchemaLink` from `apollo-link-schema`.
Good job BTW, :+1: for me except failing test

langpavel:
@nbclark Do you know what? `await` is not allowed in (arrow) functions not marked as `async`.
Please rewrite your code correctly. I'm pretty sure that this code cannot work, even cannot compile

nbclark:
Ahh yeah my oversight - Ported this in from our production code. I'll clean it up and ping you.

langpavel:
Thank you @nbclark. Can you confirm that this really works? Did you try build it in production mode and test it?

nbclark:
Hey @langpavel - I've run it in hot-reload and in release mode and it works fine. the homepage hits the new graphql query, and I've confirmed it hits internally rather than back out through the network. Tests pass as well.

We're using essentially the same code in our production site (divvyhomes.com).

nbclark:
My pleasure - Thanks for creating & maintaining this project

langpavel:
@koistya, @frenzzy, LGTM, review please. Thanks! :-)

langpavel:
@koistya, @frenzzy I like this, can you give your :white_check_mark: ?

frenzzy:
Sorry guys, I don't use `graphql` in my own projects yet and do not have much experience with it. So, I can't give you feedback here or for other gql related issues. @langpavel please decide yourself :)

PR #1512, opened 2018-01-18
Add a runtime warning when NODE_ENV is missing in release runs

jorrit:
I accidently forgot to set `NODE_ENV` to `"production"` in my production environment. React-Starter-Kit uses DefinePlugin to replace `NODE_ENV` with `"production"` at build time, but this only applies to the app code. Code like React looks at the present value of `NODE_ENV` to determine whether to load debug or production code. express also 

I propose that `BannerPlugin` is extended to issue a warning when server.js is started without `NODE_ENV` set to `"production"` when server.js is generated using `yarn build --release`.

### DIFF
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -461,7 +461,12 @@ const serverConfig = {
     // Adds a banner to the top of each generated chunk
     // https://webpack.js.org/plugins/banner-plugin/
     new webpack.BannerPlugin({
-      banner: 'require("source-map-support").install();',
+      banner: [
+        'require("source-map-support").install();',
+        !isDebug
+          ? 'if(process.env.NODE_ENV !== "production") { console.warn("You\'re running a release build without the NODE_ENV environment variable set to \'production\'.\\nThis will lead to suboptimal performance."); }'
+          : '',
+      ].join('\n'),
       raw: true,
       entryOnly: false,
     }),


jorrit:
Is there anything I can add to this PR?

frenzzy:
I think if you really want to add such warning you must do it explicitly ie in `src/server.js`

jorrit:
Isn't it harder to know whether NODE_ENV is set to a specific value in server.js, because webpack/defineplugin will resolve if statements involving NODE_ENV at compile time?

langpavel:
@frenzzy I think @jorrit is right. Webpack will replace code in server.js so this kind of check is useless then. Note that NODE_ENV is important at server in production env for external libraries.

frenzzy:
Okey, the next question is why webpack should replace `process.env.NODE_ENV` to something static for server? Maybe it should not?

jorrit:
I think currently it is done out of habit. It is more or less the default recommended configuration of webpack. I wonder if webpack 4 is going to do this by default with the new production mode. You're right, file size doesn't matter that much server side and I think those few if statements that are affected aren't a performance problem, so the optimization is not of much use.

langpavel:
@frenzzy, @jorrit`process.env` is not native javascript object but something like proxy to system environment, so accessing it often is very slow.It is not big deal to access it once in every module which need the value.
May be we can use`BannerPlugin` and`DefinePlugin` in this way on server:

```js
    // Define free variables
    // https://webpack.js.org/plugins/define-plugin/
    new webpack.DefinePlugin({
      'process.env.NODE_ENV': '__NODE_ENV', // ← This is new
      'process.env.BROWSER': true,
      __DEV__: isDebug,
    }),

    // ...

    // https://webpack.js.org/plugins/banner-plugin/
    new webpack.BannerPlugin({
      banner: [
        'require("source-map-support").install();',
        'var __NODE_ENV = process.env.NODE_ENV;', // ← This is new
        !isDebug ? 'if(__NODE_ENV !== "production") { console.warn("You\'re running a release build without the NODE_ENV environment variable set to \'production\'.\\nThis will lead to suboptimal performance."); }'
        : '',
      ].join('\n'),
      raw: true,
      entryOnly: false,
    }),
```

jorrit:
As far as I know, my patch would only add code that is executed once, at the start of the process. I don't see the benefit of the added variable.

frenzzy:
Dependencies mentioned above will use `process.env` at runtime anyway. So, we need to replace `process.env.NODE_ENV` in all dependencies (probably bundle them too) or do not touch `process.env` for server bundle at all.

frenzzy:
The final goal is to not use webpack for server at all. If you really want this check you should put it to server entry point.

huangyingwen:
server.js
`import configureStore from './store/configureStore';
import { setRuntimeVariable } from './actions/runtime';
import config from './config';

process.env = { ...process.env, NODE_ENV: process.env.NODE_ENV };

const app = express();

//
// Tell any CSS tooling (such as Material UI) to use all vendor prefixes if the
// user agent is not known.
global.navigator = global.navigator || {};
global.navigator.userAgent = global.navigator.userAgent || 'all';`

langpavel:
@huangyingwen Ok, when you do this, you effectively set build value into runtime.
Big one disadvantage here is that you can do this after all imports, so you end with half application using development environment and the half later production one.

huangyingwen:
@langpavel I do not quite understand what you mean

langpavel:
@huangyingwen This code
```js
// 23 import statements like this
import { store, actions, database, whatever } from 'whatever-package';
process.env = { ...process.env, NODE_ENV: process.env.NODE_ENV };
```
will be transformed by webpack into this:
```js
// 23 require() calls
var _ = require('whatever-package');
process.env = { ...process.env, NODE_ENV: 'production' }; // <-- Too late!! :-(
```
So you will update runtime environment to value set at build time, but **too late**, because all modules are already imported and initialized.

huangyingwen:
@langpavel Thank you, I've now understood

langpavel:
@jorrit You can add check to `server.js` like this:
```js
// `process.env.NODE_ENV` cannot be used here, you must use `process.env['NODE_ENV']`
if (!__DEV__ && process.env['NODE_ENV'] !== 'production') {
  console.warn("You\'re running a release build without the NODE_ENV environment variable set to \'production\'.\nThis will lead to suboptimal performance.");
}
```
Please test if this works and tell us! Thanks!

jorrit:
The webpack 4 changelog reads:

> process.env.NODE_ENV are set to production or development (only in built code, not in config)

So maybe this change is no longer needed after upgrading to webpack 4.

langpavel:
@jorrit This is not about Webpack 4, but about DefinePlugin here: https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L304..L308

ulani:
@jorrit  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1543, opened 2018-02-16
Add prefixUrl to package.json

wisekaa03:


rodeyseijkens:
We are having this in the [config.js](https://github.com/rodeyseijkens/react-starter-kit/blob/master/src/config.js) instead of the package.json, so we can have it running on different environments with different sub-routes if needed.

```js
  // Node.js app
  port: process.env.PORT || 3000,
  // prefix the baseurl with a '/' and leave the trailing slash out
  baseUrl: process.env.BASE_URL || '',
  // https://expressjs.com/en/guide/behind-proxies.html
  trustProxy: process.env.TRUST_PROXY || 'loopback',
```

PR #1547, opened 2018-02-23
fix babel core usage

reisel:
babel core supports a transformFileSync method which this process would highly benefit from

### DIFF
--- a/tools/messages.js
+++ b/tools/messages.js
@@ -9,7 +9,7 @@
 
 import chokidar from 'chokidar';
 import Promise from 'bluebird';
-import { transform } from 'babel-core';
+import { transformFileSync } from 'babel-core';
 import { readFile, writeFile, glob } from './lib/fs';
 import babelrc from '../.babelrc';
 import { locales } from '../src/config';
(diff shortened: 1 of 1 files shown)


langpavel:
What benefits will have sync version over async?

reisel:
this is only on the messages process. this process doesn't apply in dev hot mode.
So this process just simplifies the messages tool.

langpavel:
Oh, got it, thank you very much @reisel, merged

I'm also thinking about moving `messages` dir one level up — to the project root. Do you think that this is good idea?

Next think is new build step I'm thinking about — split messages by chunk and language
This way it should be much more usable directly with webpack, possibly move localization loading to build process instead of runtime as it is handled now...

reisel:
@langpavel hey,
I think moving them to the root folder will be a good choice, they are not really part of the source code.
regarding the splitting to chunks could improve load time, especially for apps with a lot of text.

I'm currently in the process of designing how to add the translations into our workflow.
I think I would use a google spreadsheet divided by source file with all the messages and translation and load them on compile time to build the jsons.

How would you implement the localizations in compile time?
create a different chunk for each supported language?

I really liked the approach of the md files for large texts.

Issue #1549, opened 2018-02-26
Upgrade to Webpack 4.0


rodeyseijkens:
Well [Webpack 4.0](https://github.com/webpack/webpack/releases/tag/v4.0.0) is out and it is of course awesome, so we need an upgrade.

Things to ook out for:
- You have to choose (mode or --mode) between two modes now: production or development
- ModuleConcatenationPlugin -> optimization.concatenateModules (on by default in production mode)
- CommonsChunkPlugin was removed -> optimization.splitChunks, optimization.runtimeChunk
- Resolving can now be configured with module.rules[].resolve. It's merged with the global configuration.

There is probably more but these are the first things that I saw that needed to be addressed I think.

UPDATE:
So far I have it working, but there is a weird bug in BundleAnalyzerPlugin where it cannot analyse the "shared" bundle. Not sure why but probably has to do something with JSONP
I'll make an PR soon with the "bug" in there.

joemewes:
where are you specifying the `--mode` option ?  Inline with the `babel-node tools/run start` command?  or in the build/start scripts somewhere?

rodeyseijkens:
@joemewes In the webpack config itself. 
The PR was wrong, too many changes from my own fork, but the initial setup was there:
https://github.com/rodeyseijkens/react-starter-kit/blob/73c445017ef57414dd6c9ed2bac862ccb4436488/tools/webpack.config.js#L35

tim-soft:
Have you tested BundleAnalyzerPlugin after this PR? https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/159

PR #1562, opened 2018-03-06
[Feature] Webpack 4 upgrade

rodeyseijkens:
- Updated for webpack 4
- Upgraded general packages

Home chunk is now handled by webpack and it generates the best chunks together.
Each route has it own chunk now instead of the `webpackMode: 'eager'`.
Webpack for example creates a "home\~about" and a "privacy\~login" chunk and the `server.js` pushes the corresponding chunks to the SSR `Html.js` depending on the route it has. 
This also eliminates the need for an array in the `route.chunks`, so this is renamed to `route.chunk` and a string now.

**About the WIP part**
There is still work that need needs to be done for the upcoming _"DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead"_. So in the `start.js` we use the `compiler.plugin('compile')` for example these still need to be changed to the new API.

**_If any of you have questions/enhancements please leave a comment_**
#1549


tim-soft:
How about side-effects of static render -- anything need to change here?
https://github.com/kriasoft/react-starter-kit/blob/master/tools/render.js

rodeyseijkens:
Shouldn’t be any different that before, so not affected.

rodeyseijkens:
There is still a `DeprecationWarning` but that might be of one of the plugins, but not entirely sure.

`postcss-loader` loads for some reason `schema-utils@0.4.3` instead of `0.4.5` which might be a fix but not 100% sure. `"postcss-loader > schema-utils@0.4.3" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0"`


@tim-soft On the render question, it is broken, but I what I have seen is that it is even broken on the current master branch so this might need a new PR/Issue for a fix for that.

langpavel:
> I have seen is that it is even broken on the current master branch so this might need a new PR/Issue for a fix for that.

Can you describe really quickly what is broken?

rodeyseijkens:
@langpavel sorry I thought the `yarn render` also created a build, but it isn't that was the confusion on my end. Forgot about the `yarn build --release --static`.

langpavel:
@rodeyseijkens Yeah, this should be done some way better :grimacing: I was surprises by this too :slightly_smiling_face:

rodeyseijkens:
@langpavel new package.json script maybe?
`"build-render": "yarn build --release --static"`

rodeyseijkens:
Latest commit removed `NoEmitOnErrorsPlugin` and `NamedModulesPlugin` cause its not needed or included in dev mode by default see: https://github.com/webpack/webpack/releases

I also pinned down the latest `DeprecationWarning: Tapable.plugin`, it is not in the RSK code. Its part of the [AssetsPlugin](https://github.com/kossnocorp/assets-webpack-plugin). There is an [issue open](https://github.com/kossnocorp/assets-webpack-plugin/issues/86) but I'm not entirely sure how fast they are going to fix it... cause it's been a year since an update on the package.

So what will be our status on this with this current PR?

tim-soft:
Unfortunately this is the reality for a lot of webpack plugins right now, extract text plugin is keeping me from upgrading on a lot of projects.

Maybe there is an alternative that is v4 friendly such as https://github.com/webdeveric/webpack-assets-manifest

rodeyseijkens:
I implemented the [webpack-assets-manifest](https://github.com/webdeveric/webpack-assets-manifest) as suggested by @tim-soft 

So everything is now up to date for Webpack 4.0, so in my optinion it would be ready for merge.
@langpavel @koistya @frenzzy

src/server.js
```diff
@@ -162,8 +162,12 @@ app.get('*', async (req, res, next) => {
...
     data.styles = [{ id: 'css', cssText: [...css].join('') }];
     data.scripts = [assets.vendor.js];
-    if (route.chunks) {
-      data.scripts.push(...route.chunks.map(chunk => assets[chunk].js));
+    if (route.chunk) {
+      data.scripts.push(
+        ...Object.keys(assets)
+          .filter(asset => asset.includes(route.chunk))
+          .map(chunk => assets[chunk].js),
+      );
```
langpavel: @rodeyseijkens Why did you remove support for `route.chunks`?
May be wi can support both, `chunk` and `chunks` to keep this backward compatible?

rodeyseijkens:
@langpavel the reason is that Webpack is now intelligent enough to split everything up in chunks that overlap thus you don't need to manually configure / figure out what chunks you want to combine. That is also why the 'eager' feature was dropped cause now it is exactly serving what is needed for that page and thus not any overloading of duplicate chunk data.

If we are going to manually put in more chunks it won't work because Webpack itself now figures out what chunks overlap and needs to be separate.

So for example you have a `home`, `login`, `admin` routes.
`home.js` will have its own chunk.
`login.js` will have its own chunk.
`admin.js` will have its own chunk.
`home~login~admin.js` is a shared chunk this will be generated by webpack itself cause it for example sees that a lot of componens or functions are shared (header component for example).
`login~admin.js` also ia a shared chunk and will be generated by webpack (input field component for example).
`vendors~login~admin.js` is a special vendor chunk specific for those pages (react-forms package that is only used on those pages).

Hope this clears it up.

tim-soft:
Good catch! Is there a benefit to maintaining separate client and vendor bundles?

rodeyseijkens:
@tim-soft Well kind of right I guess, maybe we don't need to have a special configuration then for Webpack splitChunks.

langpavel:
@rodeyseijkens 
> If we are going to manually put in more chunks it won't work because Webpack itself now figures out what chunks overlap and needs to be separate.

I think you are wrong (you may be right but you should proof to me :slightly_smiling_face:)
Loading multiple chunks can be implemented and should work IMHO

It is possible to have nested routes.
You can imagine that like *page* which have some *tabs* inside, like `/admin/dashboard` and `/admin/articles`.
Now you need `admin` chunk but you also need `dashboard` or `articles` chunk.
How you can do this with only one chunk? (Yes, I know, there can be potentially better way to do this, but not always I think)

langpavel:
Merged :tada:

PR #1576, opened 2018-03-24
Run build as last step on Travis CI

langpavel:


langpavel:
@frenzzy What do you think? Will this be slow?

langpavel:
It's ok, adds ~ 20 sec but will help much more with PRs

Issue #1580, opened 2018-03-25
[feature/react-intl] problem on IE10 (polyfill)


dreisel:
when running the app on IE10 (or any browser that need the intl polyfill)
the client loader will eventually load the polyfill but the main client file that uses the intl is loaded first.
this causes an error and the entire app won't load.

I fixed this issue at my fork by importing the client file dynamically when the polifilled is done (if needed).

I'll open a PR soon

langpavel:
I have this in clientLoader.js but it's possible that it was lost

ulani:
@dreisel   thank you very much for crating this issue! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

NOTE: The `main` branch has been updated with React Starter Kit v2, using JAM-style architecture.

PR #1583, opened 2018-03-26
bug(polyfill) - fix polyfill loading.

reisel:
fix bug when loading the app in an environment without native intl.
more info - #1580


langpavel:
Thank you and sorry for delay. Next time you can mention me 🙂

PR #1595, opened 2018-04-02
Update packages

mmarkelov:
Update packages and fix "Deprecation warning for String based operators"
https://github.com/sequelize/sequelize/issues/8417


jorrit:
react is at 16.3.1 now. Can you update the PR?

langpavel:
Thank you @mmarkelov

damiangreen:
This commit has broken the react-intl branch. It can no longer be merged because of conflicts in package.json and yarn.lock. I imagine that branch will also need updating.

src/data/sequelize.js
```diff
@@ -7,10 +7,11 @@
  * LICENSE.txt file in the root directory of this source tree.
  */
 
-import Sequelize from 'sequelize';
+import Sequelize, { Op } from 'sequelize';
 import config from '../config';
 
 const sequelize = new Sequelize(config.databaseUrl, {
+  operatorsAliases: Op,
```
SystemDisc: This should be set to `false` and not `Op`.

PR #1616, opened 2018-05-16
Add keywords to initial render

mmarkelov:
I suppose that keyword should be added


jorrit:
Meta keywords are 90's technology, AFAIK

PR #1647, opened 2018-07-22
Add build/start scripts for Heroku

josephfrazier:
This should make it possible to simply push to Heroku and have it
build/run the app, instead of needing to run `yarn deploy` manually.

Similarly, things like review apps and auto-deploy from master should
also work.

https://devcenter.heroku.com/articles/nodejs-support#heroku-specific-build-steps
https://devcenter.heroku.com/articles/nodejs-support#default-web-process-type
https://devcenter.heroku.com/articles/github-integration

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: node build/server.js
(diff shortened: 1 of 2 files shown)


piglovesyou:
I think having these in master branch is too specific for Heroku users.

josephfrazier:
That's fair, just thought I'd throw it out there. Thanks for all your work on this!

Issue #1626, opened 2018-06-13
when updating Babel-preset-env to 7.0.0-beta.51 forceAllTransforms option is invalid


jackrussell1996:
As of Babel-preset-env 7.0.0-beta.51, there has been additional validation on the settings that react-starter-kit webpack.config.js fails to validate against and so the module build fails:
`Module build failed: Error: [BABEL] my-project\src\client.js: Invalid Option: 'false' is
not a valid value for 'targets.forceAllTransforms'`

Related issue https://github.com/babel/babel/issues/8156

According to the documentation at https://babeljs.io/docs/en/next/babel-preset-env.html#forcealltransforms

The **forceAllTransforms** property should be set in the **root** of the preset settings object:
```
module.exports = {
  presets: [
    ["@babel/preset-env", {
      targets: {
        chrome: 59,
        edge: 13,
        firefox: 50,
      },
      forceAllTransforms: process.env === "production"
    }],
  ],
};
```
whereas in react-starter-kit it is set as a property of targets:
```
module.exports = {
  presets: [
    ["@babel/preset-env", {
      targets: {
        browsers: pkg.browserslist,
        forceAllTransforms: process.env === "production"
      },
      modules: false,
      useBuiltIns: false,
      debug: false,
    }],
  ],
};
```

I don't think that the forceAllTransforms option is being respected even when using babel/preset-env 7.0.0-beta.44

piglovesyou:
I believe https://github.com/kriasoft/react-starter-kit/pull/1673 solved this issue. Please reopen it if it didn't.

Issue #1607, opened 2018-05-03
Issue with babel preset-stage-2 


Bogdaan:
```
ERROR in ./tools/lib/webpackHotDgeevClient.js
Module build failed: Error: [BABEL] rsk/tools/lib/webpackHotDevClient.js: The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2 (While processing: "rsk/node_modules/@babel/preset-stage-2/lib/index.js")
    at rsk/node_modules/@babel/preset-stage-2/lib/index.js:107:11
    at rsk/node_modules/@babel/helper-plugin-utils/lib/index.js:18:12
    at rsk/node_modules/@babel/core/lib/config/full.js:172:14
    at cachedFunction (rsk/node_modules/@babel/core/lib/config/caching.js:42:17)
    at loadPresetDescriptor (rsk/node_modules/@babel/core/lib/config/full.js:243:63)
    at rsk/node_modules/@babel/core/lib/config/full.js:68:19
    at Array.map (<anonymous>)
    at recurseDescriptors (rsk/node_modules/@babel/core/lib/config/full.js:66:36)
    at loadFullConfig (rsk/node_modules/@babel/core/lib/config/full.js:112:6)
    at rsk/node_modules/@babel/core/lib/transform.js:28:33
 @ multi @babel/polyfill ./tools/lib/webpackHotDevClient ./src/client.js
```

Issue: https://github.com/babel/babel/issues/7786

Bogdaan:
How to reproduce:
```bash
git clone -o react-starter-kit -b master --single-branch https://github.com/kriasoft/react-starter-kit.git MyApp && \
cd MyApp &&\ 
yarn install &&\ 
yarn start
```

tim-soft:
roll back your babel libs to beta.44, so remove the `^`'s from `package.json` and reinstall packages
`"@babel/polyfill": "7.0.0-beta.44",` etc

piglovesyou:
I believe https://github.com/kriasoft/react-starter-kit/pull/1673 solved this issue. Please reopen it if it didn't .

PR #1673, opened 2018-09-22
Upgrade all dependencies including Babel 7

### LINKED ISSUE #1654: As of v7.0.0-beta.55, we've removed Babel's Stage presets.
はじめて投稿します。

react-redux-material-starter-kit-master.zip
↑をnpm run build すると　↓のエラーがでます。回避方法をご教示くださいませ。

As of v7.0.0-beta.55, we've removed Babel's Stage presets.
Please consider reading our blog post on this decision at
https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets
for more details. TL;DR is that it's more beneficial in the
  long run to explicitly add which proposals to use.

For a more automatic migration, we have updated babel-upgrade,
https://github.com/babel/babel-upgrade to do this for you with
"npx babel-upgrade".

If you want the same configuration as before:

{
  "plugins": [
    // Stage 2
    ["@babel/plugin-proposal-decorators", { "legacy": true }],
    "@babel/plugin-proposal-function-sent",
    "@babel/plugin-proposal-export-namespace-from",
    "@babel/plugin-proposal-numeric-separator",
    "@babel/plugin-proposal-throw-expressions",

    // Stage 3
    "@babel/plugin-syntax-dynamic-import",
    "@babel/plugin-syntax-import-meta",
    ["@babel/plugin-proposal-class-properties", { "loose": false }],
    "@babel/plugin-proposal-json-strings"
  ]
}

If you're using the same configuration across many separate projects,
keep in mind that you can also create your own custom presets with
whichever plugins and presets you're looking to use.

module.exports = function() {
  return {
    plugins: [
      require("@babel/plugin-syntax-dynamic-import"),
      [require("@babel/plugin-proposal-decorators"), { "legacy": true }],
      [require("@babel/plugin-proposal-class-properties"), { "loose": false }],
    ],
    presets: [
      // ...
    ],
  };
};

piglovesyou:
1. ncu --upgradeAll
4. Leave "graphql" to be "^13.2.0" that apollo packages depend on
2. Apply patch of "npx babel-upgrade" output
3. Leave only necessary babel plugins
4. Pass "yarn fix" by disabling stricter rules

fixes #1654 #1626 #1607


jorrit:
Thanks. I think it is also necessary to set `configFile: false,` in webpack.config.js in addition to `babelrc: false` to make sure only the settings in webpack.config.js are read and not the ones in babelrc.config.js.

piglovesyou:
@jorrit Thank you for reviewing, I added it.

PR #1676, opened 2018-10-02
Using fragment instead of the div tag

ali-master:
Hi there,

Since we are using the latest version of React.js then we can use the Fragment feature from now on. And I think it will be helped other users for using this awesome feature in their codes.


ali-master:
First question:
No, but I can update it asap.

Second question:
Yes I will update unit tests

piglovesyou:
This issue is one of the reasons working on #1797, thank you.

PR #1678, opened 2018-10-03
Update Layout.test.js.snap

ali-master:
This update related to pull request #1676


langpavel:
If related and directly keen on #1676, I think that it should be part of #1676. I'm wrong?

ali-master:
Yes, you're right.

piglovesyou:
I don't know why the test passed when Layout.test.js.snap was wrong at the moment, but I did this in #1797. Thank you.

PR #1697, opened 2018-12-20
Add some optimizations to the production environment

luffyZh:
First of all, I've been using `react-starter-kit` for project development always. I think it's very nice tool for enterprise application development.
The pull request solves some of the problems I encountered when deploying the project.
 - Beautify production code
   The UglifyJS is not support ES6 with Webpack 4.x. So it can be replaced by `terser-webpack-plugin`.
```
const TerserPlugin = require('terser-webpack-plugin');
module.exports = {
  //...
  optimization: {
    minimizer: [new TerserPlugin()]
  }
};
```
  Another point for beautify production code is css. `optimize-css-assets-webpack-plugin`.

 - Add procution command for auto deploy such as pm2.
   When I use pm2 to do automatic deployment on the server, the command of `yarn build-stats` because the command call the `new BundleAnalyzerPlugin()`, the console will stop and need `Ctrl + c` to close it so can't continue auto deploy. 
image
So I add a auto deploy command `build-prod`.


package.json
```diff
@@ -155,6 +157,7 @@
     "copy": "babel-node tools/run copy",
     "bundle": "babel-node tools/run bundle",
     "build": "babel-node tools/run build",
+    "build-prod": "yarn run build --release",
```
samkelleher: The `build-prod` command isn't actually needed as you're supposed to run `yarn build --release` to get a production build.

tools/webpack.config.js
```diff
@@ -368,6 +403,22 @@ const clientConfig = {
...
+            terserOptions: {
+              ecma: 6,
+              warnings: false,
+              extractComments: false,
+              compress: {
+                drop_console: true, // remove console
+              },
+              ie8: false,
+            },
+          }),
```
piglovesyou: Please allow me to ask. From my understanding of [the docs of webpack4](https://webpack.js.org/configuration/optimization/#optimizationminimizer), it minifies JavaScript by default on `production mode` with TerserPlugin. So you did it in order to overwrite [the default terser option](https://github.com/terser/terser#minify-options), correct?

luffyZh: emm, add these code for remove console && some other info from production. The default config don't do it.

piglovesyou: @luffyZh Right, thank you. Yeah so it's a good example to specify the minifier options.

Issue #378, opened 2016-01-05
Testing: TypeError: Cannot read property 'apply' of undefined at StyledComponent.componentWillMount 


flysonic10:
After setting up tests with Jest and getting through the auto-mocking issues described in #264, I'm getting:

`TypeError: Cannot read property 'apply' of undefined
        at StyledComponent.componentWillMount (decorators/withStyles.js:19:50)`

The specific line in the withStyles decorator causing the error is: 

`this.removeCss = this.context.insertCss.apply(undefined, styles);`

insertCss is undefined. How can the context be passed into the decorator for testing?
##

flysonic10:
For now, I patched it with

```
import emptyFunction from '../../node_modules/fbjs/lib/emptyFunction';

...

    const insertCss = this.context.insertCss || emptyFunction;
    this.removeCss = insertCss.apply(undefined, styles);```

**Comment by Deepblue129** [comment]
+1

**Comment by bhargavigundaa** [comment]
@flysonic10 : Can you please elaborate your answer?

**Comment by bhargavigundaa** [comment]
Beside exporting the component with withStyles decorator, we need to export it as a stand alone.This will solve the issue

export class GridView extends Component {
}

export default withStyles(s)(GridView);

And in test component use standalone one 

import GridView from './GridView';

and in other cases use default one

import { GridView } from './GridView';

**Comment by langpavel** [comment]
I think this can be closed.

**Comment by stujo** [comment]
This worked for me:


```
describe('Matrix', () => {

  it('renders numbers', () => {
    const wrapper = render(
      <App context={{ insertCss: () => {} }}>
        <Matrix values={[[1,2,3,4],[5,6,7,8],[9,10,11,12]]}>
        </Matrix>
      </App>
    );
    console.log(wrapper.html())
    expect(wrapper.find('div.child').length).to.eq(1);
  });
});
```

pedro-teixeira:
@bhargavigundaa your solution seems to be the best one to send the raw component to the Storybook, but I can't make the css work without the `insertCss`.

Any special configuration I'm missing it?

turnermatthewk:
@bhargavigundaa's approach works for me fine on classes, but I on functional components where I `export const GridView = ({ foo, bar }) => ` 

I still run into the same apply of undefined.

markopavlovic:
does anyone have a working Storybook example with this boilerplate?

timhuff:
I resolved this in the following way:
[...]

PR #1699, opened 2018-12-26
Jest.config: added a mock higher order component for isomorphic-style

M-Izadmehr:
Whenever the higher order component **isomorphic-style-loader** is used to inject styles to a component, the main default export cannot be imported with enzyme. A solution to this is to only import the base component (and not the default component). This solves some of the problems. 

However, if we try to **mount** a component (instead of **shallow** rendering), and if any of the children use withStyles from isomorphic-style-loader, we again face the same problem. As a result, a lot of the scenarios fail.

This has been an issue, which was reported several times:

[Issue](https://github.com/kriasoft/react-starter-kit/issues/378)
[Issue](https://github.com/kriasoft/isomorphic-style-loader/issues/130)

There is a straightforward solution to solve this concern (which was also discussed in one of issues), I also wrote a [medium](https://medium.com/@m.izadmehr/react-starter-kit-testing-isomorphic-style-loader-with-jest-and-enzyme-fd361b0591f0) article discussing the details of the changes:

Inside jest.config.js, there is an option called moduleNameMapper, which can be used to stub-out resources or mock them. And we add a field, to mock all instances of imports of isomorphic-style-loader, with a simple mock function.

I think it is a great idea to add the required changes in the jest.config.js file by default.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -62,8 +62,6 @@ module.exports = {
   // A map from regular expressions to module names that allow to stub out resources,
   // like images or styles with a single module.
   moduleNameMapper: {
-    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
-      '<rootDir>/tools/mocks//staticAssetMock.js',
     '\\.(css|less|sass|scss)$': 'identity-obj-proxy',
     'isomorphic-style-loader/lib/withStyles':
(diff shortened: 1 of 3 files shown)


ulani:
@M-Izadmehr  thank you very much for this PR! Unfortunately, we have close it due to inactivity. Feel free to re-open it or [join](https://discord.com/invite/2nKEnKq) our [Discord](https://discord.com/invite/2nKEnKq) channel for [discussion](https://github.com/kriasoft/react-starter-kit/discussions/1950).

PR #1721, opened 2019-04-01
Upgrade Dockerfile with multi-stage build

huynhminhtufu:
The old Dockerfile has an error, at line 13:

```
COPY ./build .
```

This command copy `build` dir from **host machine** (not from Docker build result), which is weird?!

I do some changes to optimize the docker build progress, to optimize the production image size.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+build
+node_modules
+docs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,24 @@
-FROM node:8.10.0-alpine
+FROM node:8.10.0-alpine as builder
 
 # Set a working directory
 WORKDIR /usr/src/app
 
-COPY ./build/package.json .
-COPY ./build/yarn.lock .
+COPY . .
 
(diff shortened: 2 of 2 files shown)


huynhminhtufu:
More about the error at line 13, if `build` dir doesn't exist on the sourcecode (of host machine), then the **docker build process will print an error**:

```
COPY failed: stat /var/lib/docker/tmp/docker-builder883685428/build/package.json: no such file or directory
```

Because: `./build/package.json` is not exist.

This mean use must to run `yarn build` before the Docker build, which is so weird?!

piglovesyou:
Hi @huynhminhtufu . Instead of running `docker` command, it is supposed to run `yarn build --release --docker` to pack a docker image from build result on your local machine. I would like to suggest reading [how-to-build-test-deploy](https://github.com/kriasoft/react-starter-kit/blob/master/docs/getting-started.md#how-to-build-test-deploy) for the information.

piglovesyou:
I believe the change is not necessary here. Please reopen this if you have further questions.

PR #1726, opened 2019-04-16
Fix styles error in Home.css and deprecation warnings

kutaloweb:
Here should be a valid CSS instead of SCSS so that Webstorm doesn't show an error. The second commit fixes console warnings: history/createBrowserHistory deprecated and history/PathUtils deprecated.

### DIFF
--- a/src/routes/home/Home.css
+++ b/src/routes/home/Home.css
@@ -28,22 +28,20 @@
   font-size: 1.5rem;
 }
 
-.newsDesc {
-  h1,
-  h2,
-  h3,
-  h4,
-  h5,
-  h6 {
-    font-size: 1.125rem;
-  }
+.newsDesc h1,
+.newsDesc h2,
+.newsDesc h3,
+.newsDesc h4,
+.newsDesc h5,
+.newsDesc h6 {
+  font-size: 1.125rem;
+}
 
-  pre {
-    white-space: pre-wrap;
-    font-size: 0.875rem;
-  }
+.newsDesc pre {
+  white-space: pre-wrap;
+  font-size: 0.875rem;
+}
 
-  img {
-    max-width: 100%;
-  }
+.newsDesc img {
+  max-width: 100%;
 }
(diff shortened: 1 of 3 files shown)


piglovesyou:
@kutaloweb Nice. The change of `history/PathUtils ` seems necessary, but `home.css` fix seems not, since [CSS Nesting](https://cssdb.org/#nesting-rules) is one of valid proposals in cssdb (used by PostCSS). I think we should not drop features depending on whether VSCode supports it, what do you think?

Or, we can discuss from which stage of [cssdb.org](https://cssdb.org/#staging-process) react-starter-kit supports by using [postcss-preset-env](https://preset-env.cssdb.org/features). What do you think @langpavel ?

langpavel:
I have no opinion on css @frenzzy? LGTM.

src/routes/home/Home.css
```diff
@@ -28,22 +28,20 @@
...
+  font-size: 1.125rem;
+}
+
+.newsDesc pre {
+  white-space: pre-wrap;
+  font-size: 0.875rem;
+}
+
+.newsDesc img {
+  max-width: 100%;
```
langpavel: I slightly more like the old version of CSS

piglovesyou: Please consider the author of the PR doesn't notice [the postcss plugin enables CSS nesting](https://github.com/kriasoft/react-starter-kit/blob/8d6c018f3198dec2a580ecafb011a32f06e38dbf/tools/postcss.config.js#L41). 

But problem is CSS Nesting is far from standar

kutaloweb:
@langpavel @piglovesyou I understand what are you talking about but is very annoying to see the errors for developers who use WebStorm (home, routs, src and the whole project directories are marked with red color). I don't know if other IDEs have the same issue. I change these styles every time I start a new project. I have no choice.

piglovesyou:
It's not only about `css-nesting`. Once we introduce `postcss-preset-env` and choose suppose `stage 3+`, your concern will disappear so please hold on.

piglovesyou:
Thank you @kutaloweb . I'll work on `postcss-preset-env` things later on  #1731 .

Issue #1529, opened 2018-02-04
Any Mocha docs need to be erased


piglovesyou:
From #657 . There is already Jest installed so any docs should not include something about Mocha to avoid confusion.

> @langpavel
> Documentation indeed needs improvements

There are still article written about Mocha:
https://github.com/kriasoft/react-starter-kit/blob/master/docs/testing-your-application.md#testing-your-application

langpavel:
Good point. This should be easy PR! :-)

aguscha333:
I can work on this if it's not taken already 😄

aguscha333:
Wasn't this done already in #1624?

piglovesyou:
@aguscha333 Seems you're right. And I found a few more "mocha" descriptions left in `docs/getting-started.md` so glad if you would modify that part too.

aguscha333:
@piglovesyou done, link to the PR #1729

piglovesyou:
Resolved by #1729 .

PR #1729, opened 2019-05-17
Remove mocha form docs (getting started)

aguscha333:
# Remove mocha form docs (getting started)

This PR removes the last mentions of Mocha in the docs, addressing the following issue #1529. I checked and the only other mention of Mocha in the whole project left is in the changelog.

Explanation about the naming of the files `Jest` looks for is still valid so I didn't need to change that.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -139,11 +139,11 @@ $ yarn run lint
 To launch unit tests:
 
 ```shell
-$ yarn run test          # Run unit tests with Mocha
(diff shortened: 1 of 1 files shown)


aguscha333:
Does anyone know why travis is failing in node stable test? The error I am getting is: `The command "eval yarn " failed.`. I only changed a `.md` file. That can't possibly be the problem.

piglovesyou:
@aguscha333 I think it's Node12 problem. There should be another issue to solve it 👍

PR #1730, opened 2019-05-26
Fix Node 12 build failure

piglovesyou:
`master` fails its build because of `sqlite3`. Usding the latest `sqlite3`, which supported Node 12 on [v4.0.6](https://github.com/mapbox/node-sqlite3/releases/tag/v4.0.6), fixes the problem.

* Update sqlite3
* Add Ci builds against the latest active LTS (10)

Or at this time we can remove Node 6 CI build since it's out of LTS support. What do you think? @langpavel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - 'stable'
+  - '10'
   - '8'
   - '6'
 env:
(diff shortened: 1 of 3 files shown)


piglovesyou:
I checked that It passes build, but doesn't launch in Node 12. Need more fix.

piglovesyou:
No, it was just my IDE puts `--inspect --debug-brk` on start which is deprecated in Node 12. Now I checked both `yarn build --release; node build/server.js` and `yarn start` work fine on my MacOS.

langpavel:
Thanks!

PR #1715, opened 2019-02-19
Upgraded react, react-dom and isomorphic-style-loader.

mglace:
Upgraded react, react-dom and isomorphic-style-loader packages.  Upgraded react context references to new context API.

PR #1759, opened 2019-11-02
Upgrade react, react-dom and isomorphic-style-loader

piglovesyou:
With respect, I used @mglace 's react-starter-kit change

Fixes kriasoft/react-starter-kit#1715


piglovesyou:
Mmm not sure why the Layout.test.js fails

PR #1829, opened 2019-11-22
Update getting-started.md

maapteh:
From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded


piglovesyou:
Thank you!

PR #1851, opened 2019-11-26
Use test-watch instead of test:watch

mgomezg:
According to the read me it should go:

yarn run test-watch             # Run unit tests in watch mode


piglovesyou:
Thank you!

PR #1963, opened 2022-05-08
Webpack + Dependencies Upgrade

andrewstec:
These changes should prepare you a bit for Node 17 and will likely need to happen anyways. However, I still had to downgrade to Node v16.15.0 LTS in the end because there is a proxy.ts file issue with Node 17 where the build breaks.

I manually tested on a Mac on v16.15.0 LTS. I have also battled Webpack with Node upgrades many times in the past, but any additional testing/sanity checks are super welcome. I only manually tested.

Just FYI, I am not sure what sort of backwards compatibility tests you consider, and I did not test any versions of Node older than 16.15.0.


koistya:
@andrewstec looks good!

PR #2044, opened 2023-05-26
Make onboarding even simpler for VS Code users

christian-bromann:
VS Code allows to create links to checkout the repository automatically:

more-easier

I already made a PR in the relay project similar to that 😉


christian-bromann:
Any comments?

PR #2167, opened 2026-03-12
feat(app): add light/dark/system theme switching

AchuAshwath:
### Summary

Connects the Settings appearance control to a working theme system: a three-way preference (light / dark / system), persisted, synced across tabs, and applied before first paint.

### Design decisions

**Jotai atoms rather than a React context.** The app already uses Jotai for shared state, and the effective theme is _derived_ from the preference — as a computed atom it cannot fall out of step. Both atoms are private, and `useTheme()` declares its return type explicitly so Jotai's `RESET` token and updater overloads stay out of app code.

**`preference` vs `theme`.** `preference` is what the user chose (`light | dark | system`); `theme` is what renders (`light | dark`). Only the preference is persisted, and nothing is written until the user picks one — so an untouched install keeps following the OS.

**The inline script in `index.html` duplicates resolution logic on purpose.** Nothing else runs before first paint, so a dark-mode user would otherwise get a flash of white while the bundle loads. `theme.test.tsx` pins its storage key and encoding.

**Theme colors live in CSS.** `ThemeSync` reads `--theme-color` from `globals.css` rather than holding hex values in TypeScript. `index.html` and `site.manifest` must repeat them since they load before any stylesheet — a manual contract recorded in `apps/app/AGENTS.md`, and one no test can cover, since Vitest resolves CSS imports to an empty string.

**Radix `ToggleGroup` instead of custom keyboard handling.** `type="single"` renders `role="radiogroup"` with `role="radio"` items and arrow-key navigation, so the switcher carries no keyboard code of its own.

### Worth a reviewer's attention

- `apps/app/vitest.setup.ts` installs a real `localStorage`: Node's Web Storage global shadows happy-dom's and stays undefined, which blocked every persistence test.
- Components in `packages/ui` must import each other **relatively**. The shadcn CLI emits `@/components/…`, but `@/` resolves through the _consuming_ app — that form builds in `apps/app` and fails in `apps/web`.
- Also rewrites `docs/frontend/ui.md` and adds `packages/ui/AGENTS.md` for the shadcn workflow.

### Testing

`bun test` · `bun typecheck` · `bun lint` · `bun build` · `bun docs:build` all clean.

Checked in Chromium against the production build: with a stored dark preference the page loads with `.dark`, `color-scheme: dark`, and `theme-color: #0f0f0f`, and cross-tab `storage` events update all three.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ node_modules/
 /.cache
 /*/.swc/
 .eslintcache
+tmp/
 
 # Testing
 /coverage
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,7 +40,7 @@ bun ui:add <component>         # Add shadcn/ui component to packages/ui
 - API worker has `nodejs_compat` enabled; web and app workers do NOT.
 - Web worker routes: `/api/*` → API worker, app routes → App worker, static → assets.
 - Service bindings connect workers internally (no public cross-worker URLs).
-- Database, auth, routing, and tRPC conventions are in subdirectory `AGENTS.md` files.
+- Per-workspace conventions live in subdirectory `AGENTS.md` files: `apps/api/` (tRPC, auth), `apps/app/` (routing, theming), `db/` (schema), `packages/ui/` (components).
 
 ## Design Philosophy
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 ## Monorepo Structure
 
-- `apps/web/` — Edge worker; routes traffic to app/api workers via service bindings
+- `apps/web/` — Astro marketing site, served by an edge worker that also routes traffic to the app/api workers via service bindings
 - `apps/app/` — Main SPA (React, TanStack Router file-based routing)
 - `apps/api/` — API server (Hono + tRPC + Better Auth)
 - `apps/email/` — React Email templates (built before API dev server starts)
@@ -36,11 +36,11 @@ bun ui:add <component>         # Add shadcn/ui component to packages/ui
 
 ## Architecture
 
-- Three workers: web (edge router), app (SPA assets), api (Hono server).
+- Three workers: web (marketing site + edge router), app (SPA assets), api (Hono server).
 - API worker has `nodejs_compat` enabled; web and app workers do NOT.
 - Web worker routes: `/api/*` → API worker, app routes → App worker, static → assets.
 - Service bindings connect workers internally (no public cross-worker URLs).
-- Per-workspace conventions live in subdirectory `AGENTS.md` files: `apps/api/` (tRPC, auth), `apps/app/` (routing, theming), `db/` (schema), `packages/ui/` (components).
+- Per-workspace conventions live in subdirectory `AGENTS.md` files: `apps/api/`, `apps/app/`, `db/`, `packages/ui/`.
 
 ## Design Philosophy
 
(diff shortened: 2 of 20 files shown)


koistya:
Theme switching is much closer now, but I still see a few issues worth fixing before merge.

### 1. First paint still flashes the light theme

`ThemeProvider` only applies the `dark` class in `useLayoutEffect`, which runs after React mounts. On a cold load, users with a dark preference will still get a light first paint before the class is added.

That is especially visible here because the app background/text colors come from CSS variables on `:root` / `.dark`.

Suggested fix:

- add a tiny inline script in `apps/app/index.html` that reads the stored preference before `/index.tsx` runs
- if there is no explicit preference yet, fall back to `matchMedia("(prefers-color-scheme: dark)")`
- set `document.documentElement.classList.toggle("dark", isDark)` there

### 2. The "system fallback" is immediately persisted as an explicit choice

Right now `getInitialTheme()` falls back to the OS preference, but `useLayoutEffect` then writes that resolved value straight into localStorage:

- no stored value
- OS is dark
- app initializes to `"dark"`
- effect persists `"dark"` to `app-theme`

After that first mount, the app is no longer "following system" at all; it has silently converted the OS-derived value into a fixed user preference.

So even if the user never touched the toggle:

- future visits keep using the old stored value
- OS theme changes are ignored
- there is no way back to "follow system" from the UI

I think the state needs to distinguish:

- persisted user preference: `light | dark | system`
- resolved theme applied to the DOM: `light | dark`

Then only persist the preference, and derive the resolved theme from `matchMedia` when the preference is `system`.

### 3. No listener for live OS theme changes

Related to the above: there is no `matchMedia(...).addEventListener("change", ...)` listener, so the app never updates when the OS theme changes while the app is open.

If the intended behavior is "follow system unless the user explicitly overrides it", this needs a media-query change listener in addition to storing a `system` preference.

### 4. `meta[name="theme-color"]` stays light in dark mode

`apps/app/index.html` still hardcodes:

```html
<meta name="theme-color" content="#fafafa" />
```

So browser chrome on mobile can stay light even when the app itself is dark. This should be updated alongside the root class, either:

- via a second dark/light pair of meta tags if the browser supports it, or
- by mutating the existing meta tag when the resolved theme changes

### Test gap

The new tests cover localStorage reads/writes and storage events well, but they do not cover the main behavioral gap above:

- "no stored value" should not become a permanent explicit `"light"` / `"dark"` preference on first mount
- OS theme changes should update the resolved theme when preference is `system`
- first-paint behavior is handled outside React, so that part needs coverage at the `index.html` / bootstrap level or at least explicit manual verification

If you want to keep this PR scoped, I’d suggest:

1. add a pre-paint script in `index.html`
2. store a theme preference (`light | dark | system`) instead of only a resolved theme
3. add a media-query change listener for the `system` case
4. update `theme-color` from the resolved theme

AchuAshwath:
@koistya  Added a few more things to this PR:

- **3-way preference** — light, dark, or "system" (follows OS)
- **Pre-paint script** — eliminates first-load flash by setting theme before React loads
- **Live OS listener** — app updates automatically when OS theme changes (if set to "system")
- **Transition suppression** — removes the "flash" when toggling themes
- **Accessible toggle** — swapped custom buttons for RadioGroup
- **meta theme-color sync** — mobile browser chrome now matches the theme
- **Sync comments** — added notes in both bootstrap and runtime to keep them in sync going forward

> ⚠️ One note: the transition suppression uses `* { transition: none !important; }`. Works great but is broad — if we see animation issues later, we'll scope it to color properties only.

### Quick manual check
- Cold load with OS dark mode (no stored pref) — should skip flash
- Toggle theme — should feel instant
- Set to "System", then change OS theme — app should react live

Tests are updated and all pass. Let me know if anything needs tweaking!

AchuAshwath:
@koistya kind reminder to look at this PR and so that we shall fix the Theme switcher bug, which is present for more than 2 months now.

koistya (approved):

